PR #3, opened 2023-11-29
Added pipedrive-thesis endpoints

theCoffeehound:
Requesting PR to #2 

Added pipedrive-thesis endpoints for deals and leads. Includes endpoints required to add and remove interest towards projects.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -104,7 +104,7 @@ paths:
           description: Pipedrive Deal status    
       responses:
         "200":
-          description: Lists open Deals from Pipedrive
+          description: Lists Deals from Pipedrive
           content:
             application/json;charset=utf-8:
               schema:
@@ -126,10 +126,10 @@ paths:
           required: true
           schema:
             type: string
-          description: Leas Id used to identify a pipedrive lead
+          description: Lead Id used to identify a pipedrive lead
       responses:
         "200":
-          description: Get more data of a lead 
+          description: Successful operation
           content:
             application/json;charset=utf-8:
               schema:
@@ -154,7 +154,7 @@ paths:
           description: Deal Id used to identify a pipedrive deal
       responses:
         "200":
-          description: Get more data of a deal 
+          description: Successful operation
           content:
             application/json;charset=utf-8:
               schema:
@@ -365,10 +365,10 @@ components:
           description: Title or a Name of a Lead
         interested:
           type: string
-          description: String where userIds are stored separeted with semicolon.
+          description: String where userIds are stored separated with semicolon.
         usedTech:
           type: string
-          description: String where usedTech are stored separeted with semicolon.
+          description: String where usedTech are stored separated with semicolon.
         addTime:
           type: string
           format: date-time
@@ -386,6 +386,7 @@ components:
           items:
             type: string
           description: String array where Lead labels are stored. labelIds are 'Hot', 'Warm' or 'Cold'
+
     Deal:
       type: object
       description: Deal data from Pipedrive API
@@ -411,10 +412,10 @@ components:
           description: Title or a Name of a Deal
         interested:
           type: string
-          description: String where userIds are stored separeted with semicolon.
+          description: String where userIds are stored separated with semicolon.
         usedTech:
           type: string
-          description: String where usedTech are stored separeted with semicolon.
+          description: String where usedTech are stored separated with semicolon.
         value:
           type: number
           description: Value of a deal
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -3,12 +3,11 @@ info:
   title: Timebank MyWork Forecast lambda functions
   description: Timebank MyWork Forecast lambda functions
   version: 1.0.0
-  
+
 security:
-   - bearerAuth: []
-  
+  - bearerAuth: []
+
 paths:
-  
   "/allocations":
     get:
       operationId: listAllocations
@@ -42,9 +41,9 @@ paths:
             application/json;charset=utf-8:
               schema:
                 type: array
-                items: 
+                items:
                   $ref: "#/components/schemas/Projects"
-  
+
   "/tasks/{projectId}":
     get:
       operationId: listProjectTasks
@@ -69,17 +68,180 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Tasks"
-  
-   
+
+  "/salesLeads":
+    get:
+      operationId: getAllLeads
+      summary: Lists all Leads
+      description: |
+        Lists Leads from Pipedrive
+      tags:
+        - Leads
+      responses:
+        "200":
+          description: Lists Leads from Pipedrive
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Leads"
+  "/salesDeals":
+    get:
+      operationId: getAllDeals
+      summary: Lists all open Deals
+      description: |
+        Lists Deals from Pipedrive
+      tags:
+        - Deals
+      responses:
+        "200":
+          description: Lists open Deals from Pipedrive
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Deals"
+
+  "/salesDealsWon":
+    get:
+      operationId: getAllDealswon
+      summary: Lists all won Deals
+      description: |
+        Lists won Deals from Pipedrive
+      tags:
+        - Deals
+      responses:
+        "200":
+          description: Lists won Deals from Pipedrive
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Deals"
+
+  "/getLeadOrDealById/{rowtype}/{id}":
+    get:
+      operationId: getDealById
+      summary: List any salesProject (lead or deal) by id
+      tags:
+        - LeadOrDealById
+      description: |
+        List salesproject with the id
+      parameters:
+        - name: rowtype
+          in: path
+          required: true
+          schema:
+            type: string
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: A list of all salesProjects
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/LeadOrDeal"
+
+  "/addDealInterest/{id}":
+    put:
+      summary: Add deal interest by ID
+      tags:
+        - Interest
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
+
+  "/addLeadInterest/{id}":
+    patch:
+      summary: Add lead interest by ID
+      tags:
+        - Interest
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
+
+  "/removeDealInterest/{id}":
+    put:
+      summary: Remove deal interest by ID
+      tags:
+        - Interest
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
+
+  "/removeLeadInterest/{id}":
+    patch:
+      summary: Remove lead interest by ID
+      tags:
+        - Interest
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
 
 components:
-      
   securitySchemes:
     bearerAuth:
       type: http
       scheme: bearer
       bearerFormat: JWT
-      
+
   schemas:
     Allocations:
       type: object
@@ -89,7 +251,7 @@ components:
           type: integer
         project:
           type: integer
-        person: 
+        person:
           type: integer
         startDate:
           type: string
@@ -113,12 +275,12 @@ components:
           type: integer
         notes:
           type: string
-          
+
     Projects:
       type: object
       description: Project data from Forecast API
       properties:
-        id: 
+        id:
           type: integer
         name:
           type: string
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
                   $ref: "#/components/schemas/Tasks"
-  
-   
+
+  "/salesLeads":
+    get:
+      operationId: getAllLeads
+      summary: Lists all Leads
+      description: |
+        Lists Leads from Pipedrive
```
Jdallos: Make sure all summaries and descriptions say its from Pipedrive. This description is fine but not all of the newly added endpoints descriptions say this.

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Tasks"
-  
-   
+
+  "/salesLeads":
+    get:
+      operationId: getAllLeads
```
Jdallos: Following the naming convention for the existing endpoints, could this operationId be renamed listSalesLeads

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+          description: Lists Leads from Pipedrive
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Leads"
+  "/salesDeals":
+    get:
+      operationId: getAllDeals
```
```suggestion
      operationId: listDeals
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        - Leads
+      responses:
+        "200":
+          description: Lists Leads from Pipedrive
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Leads"
```
```suggestion
                  $ref: "#/components/schemas/Leads"
                  
```

Jdallos: In case this is not clear, the suggestion is adding a line break between the salesLeads and the salesDeals endpoints.

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Deals"
+
+  "/salesDealsWon":
+    get:
+      operationId: getAllDealswon
```
```suggestion
      operationId: listDealsWon
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        "200":
+          description: Lists open Deals from Pipedrive
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Deals"
+
+  "/salesDealsWon":
```
Jdallos: From  looking at the pipedrive documentation it looks to me like this is the same endpoint in pipedrive but with a different query parameter. Rather than have a seperate lambda function and endpoint for this, could you add the the option to pass the appropriate query parameter to the listDeals functions which then passes it along to the forecast?

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        "200":
+          description: Lists won Deals from Pipedrive
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Deals"
+
+  "/getLeadOrDealById/{rowtype}/{id}":
```
Jdallos: Is this using the get details of a deal and the get one lead endpoints combined into one? I think following typical rest patterns here would be good so this should be split into 2 endpoints a salesDeals/{dealId} and a salesLeads/{leadId} endpoints and order then within this document beneath the related list endpoint

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        "200":
+          description: A list of all salesProjects
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/LeadOrDeal"
+
+  "/addDealInterest/{id}":
```
```suggestion
  "/addDealInterest/{dealId}":
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
+
+  "/addLeadInterest/{id}":
```
```suggestion
  "/addLeadInterest/{leadId}":
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+    patch:
```
Jdallos, line 190: Hmm, interesting that the pipedrive API uses put for updating a deal and patch for updating a lead, so no issue here, just a comment :)

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+                items:
+                  $ref: "#/components/schemas/LeadOrDeal"
+
+  "/addDealInterest/{id}":
+    put:
+      summary: Add deal interest by ID
+      tags:
+        - Interest
+      parameters:
+        - name: id
```
```suggestion
        - name: dealId
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+                type: array
+                items:
+                  $ref: "#/components/schemas/LeadOrDeal"
+
+  "/addDealInterest/{id}":
+    put:
+      summary: Add deal interest by ID
+      tags:
+        - Interest
+      parameters:
```
Jdallos, line 172: Parameter is missing description

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
+
+  "/removeDealInterest/{id}":
```
```suggestion
  "/removeDealInterest/{dealId}":
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+    put:
+      summary: Remove deal interest by ID
+      tags:
+        - Interest
+      parameters:
```
Jdallos, line 220: missing description

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        - name: id
```
```suggestion
        - name: dealId
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+        "200":
+          description: Successful operation
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Interests"
+
+  "/removeLeadInterest/{id}":
```
```suggestion
  "/removeLeadInterest/{leadId}":
```

swagger.yaml
```diff
@@ -69,17 +68,180 @@ paths:
...
+    patch:
+      summary: Remove lead interest by ID
+      tags:
+        - Interest
+      parameters:
```
Jdallos, line 244: Missing description

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
           type: array
-          items: 
-            type: integer
\ No newline at end of file
+          items:
+            type: integer
+
+    Leads:
+      type: object
+      description: Lead data from Pipedrive API
```
Jdallos, line 348: Before listing the properties you should define which ones are required, I suspect they are all required.You can see an example of how this is done in this spec: https://github.com/Metatavu/time-bank-api-spec/blob/e8b5222db8ad2762c5fc39cd11206ccc85a97026/swagger.yaml

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+      properties:
```
Jdallos, line 358: All properties should have a description, also any where an id us being used if it is a uuid then you should set the format as uuid, this is also the case when defining the parameters schema for the endpoints that are using ids.

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+        id:
+          type: string
+        title:
+          type: string
+        interested:
+          type: string
+          description: "Might need to be changed to Pipedrive API item-key"
```
Jdallos: What does this mean?

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+        usedTech:
+          type: string
+          description: "Might need to be changed to Pipedrive API item-key"
+        add_time:
```
Jdallos: Even if pipedrive is using snake_case you could map over the response to rename for our use using camelCase- which would be very nice, and these all these snake_case properties can be renamed

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+          type: string
+          format: date-time
+        status:
+          type: string
+        next_activity_subject:
+          type: string
+        next_activity_note:
+          type: string
+
+    LeadOrDeal:
```
Jdallos: When this endpoint is split as I described you should be able to remove this schema and use the existing deal or lead schema

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
           format: date
         assignedPersons:
           type: array
-          items: 
-            type: integer
\ No newline at end of file
+          items:
+            type: integer
+
+    Leads:
```
```suggestion
    Lead:
```

Jdallos: And you will need to rename the ref also

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+          type: string
+          format: date-time
+        next_activity_date:
+          type: string
+          format: date-time
+        label_ids:
+          type: array
+          items:
+            type: string
+    Deals:
```
```suggestion
    Deal:
```

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+          type: array
+          items:
+            type: string
+          description: "Added from Lead"
+
+    Interests:
+      type: object
+      description: Adding or removing interest to Pipedrive API
+      properties:
+        pipedrive-interest-key-here:
```
Jdallos: Can you clarify this?

swagger.yaml
```diff
@@ -155,5 +317,116 @@ components:
...
+        currency:
+          type: string
+          description: "Added from Deal"
+        label_ids:
+          type: array
+          items:
+            type: string
+          description: "Added from Lead"
+
+    Interests:
```
```suggestion
    Interest:
```

Jdallos (changes requested):

swagger.yaml
theCoffeehound: If you mean the "pipedrive-interest-key-here". The variables in PUT and PATCH send the variable name instead the value and gave me paranoia. But I think only place where the Pipedrives field-key matters is in the PUT/PATCH function.

swagger.yaml
theCoffeehound: Yep, this gave me headache :D

swagger.yaml
theCoffeehound: Same case as the pipedrive-interest-key-here. I commented more about this to the Interest schema at the bottom.

swagger.yaml
```diff
@@ -69,17 +68,203 @@ paths:
...
+        "200":
+          description: Lists Leads from Pipedrive
+          content:
+            application/json;charset=utf-8:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Lead"
+
+  "/salesDeals/{status}":
```
Jdallos, line 90: Is it necessary for status to be required here?

swagger.yaml
```diff
@@ -69,17 +68,203 @@ paths:
...
+        - LeadById
+      description: |
+        Get more data of a lead by id from Pipedrive
+      parameters:
+        - name: leadId
+          in: path
+          required: true
+          schema:
+            type: string
+          description: Leas Id used to identify a pipedrive lead
```
```suggestion
          description: Lead Id used to identify a pipedrive lead
```

swagger.yaml
```diff
@@ -69,17 +68,203 @@ paths:
...
+      parameters:
+        - name: status
+          in: path
+          required: true
+          schema:
+            type: string  
+          description: Pipedrive Deal status    
+      responses:
+        "200":
+          description: Lists open Deals from Pipedrive
```
```suggestion
          description: Lists Deals from Pipedrive
```

Jdallos: OR should this list open deals, or does it depend on the type of status passed in for the request. You could set a default status if no status is passed, or have no options to pass any status and have open as the default. Whatever you think is best but I think this needs to be clarified.

swagger.yaml
```diff
@@ -69,17 +68,203 @@ paths:
...
+      parameters:
+        - name: dealId
+          in: path
+          required: true
+          schema:
+            type: integer
+          description: Deal Id used to identify a pipedrive deal
+      responses:
+        "200":
+          description: Get more data of a deal
```
```suggestion
          description: Successful operation
```

swagger.yaml
```diff
@@ -69,17 +68,203 @@ paths:
...
+      parameters:
+        - name: leadId
+          in: path
+          required: true
+          schema:
+            type: string
+          description: Leas Id used to identify a pipedrive lead
+      responses:
+        "200":
+          description: Get more data of a lead
```
```suggestion
          description: Successful operation
```

swagger.yaml
```diff
@@ -155,5 +340,115 @@ components:
...
+        leadId:
+          type: string
+          format: uuid
+          description: LeadId is used to identify projects
+        title:
+          type: string
+          description: Title or a Name of a Lead
+        interested:
+          type: string
+          description: String where userIds are stored separeted with semicolon.
```
```suggestion
          description: String where userIds are stored separated with semicolon.
```

swagger.yaml
```diff
@@ -155,5 +340,115 @@ components:
...
+        usedTech:
+          type: string
+          description: String where usedTech are stored separeted with semicolon.
```
```suggestion
          description: String where usedTech are stored separated with semicolon.
```

swagger.yaml
```diff
@@ -155,5 +340,115 @@ components:
...
+        nextActivityDate:
+          type: string
+          format: date-time
+          description: Time when next activity is going to happen on a Lead
+        labelIds:
+          type: array
+          items:
+            type: string
+          description: String array where Lead labels are stored. labelIds are 'Hot', 'Warm' or 'Cold'
+    Deal:
```
Jdallos, line 390:
```suggestion

    Deal:
```

swagger.yaml
```diff
@@ -155,5 +340,115 @@ components:
...
+      properties:
+        dealId:
+          type: integer
+          description: DealId is used to identify projects
+        title:
+          type: string
+          description: Title or a Name of a Deal
+        interested:
+          type: string
+          description: String where userIds are stored separeted with semicolon.
```
```suggestion
          description: String where userIds are stored separated with semicolon.
```

swagger.yaml
```diff
@@ -155,5 +340,115 @@ components:
...
+        usedTech:
+          type: string
+          description: String where usedTech are stored separeted with semicolon.
```
```suggestion
          description: String where usedTech are stored separated with semicolon.
```

swagger.yaml
```diff
@@ -155,5 +340,115 @@ components:
...
+
+    Interest:
+      type: object
+      description: Adding or removing interest to Pipedrive API
+      required:
+        - interest
+      properties:
+        interest:
+          type: string
+          description: String where userIds are stored separeted with semicolon.
```
```suggestion
          description: String where userIds are stored separated with semicolon.
```

Jdallos (approved):
Approved with several suggestions which can be committed directly and a comment about the use of status as a required parameter for salesDeals

swagger.yaml
theCoffeehound: Yes for now

PR #10, opened 2024-03-25
Add slack user avatar endpoint

Kseniia-Koshkina:


swagger.yaml
```diff
@@ -172,6 +168,30 @@ paths:
...
+  "/slack-avatar":
+    get:
+      operationId: getSlackUserAvatar
+      summary: Get person slack avatar
+      description: |
+        Get slack avatar
+      tags:
+        - Slack avatar
+      parameters:
+        - name: email
```
Jdallos: Would there be someway we could only need to request the lambda once to get the images, rather than needing to request for every user? Would it be feasible to get an image for all the slack users at once and return it as an object with an id/ name?

swagger.yaml
```diff
@@ -511,6 +531,17 @@ components:
...
+    Slack-avatar:
+      type: object
+      description: Project data time entries from Forecast API
+      properties:
+        image_original:
+          type: string
+        image_24:
+          type: string
+        image_32:
+          type: string
```
Jdallos: Do we need 3 image formats? Could this response contain all of the slack users images and some way of identifying which user an image is associated with.
Not directly related to this but is it worth thinking about storing the images in local storage in the UI. Probably not, but at least it would be nice to only need to make the request once for all users rather than for each user.

Jdallos (changes requested):

Jdallos (approved):

PR #17, opened 2024-04-21
Added the software registry endpoints

### LINKED ISSUE #16: add software registry function to spec thesis work

yuusufisse:


swagger.yaml
```diff
@@ -402,6 +402,98 @@ paths:
         "200":
           description: Paid status updated
 
+  "/software":
+    get:
+      summary: Get all software entries
```
Jdallos, line 408: Your endpoints appear to be missing operationId and tags. you can see examples of these in the other endpoint described in this spec.

swagger.yaml
```diff
@@ -402,6 +402,98 @@ paths:
...
+    post:
+      summary: Register new software
+      description: Register a new software entry in the registry.
+      requestBody:
+        description: Software entry to register
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/SoftwareRegistry'
```
Jdallos: Use double quotes throughout

swagger.yaml
```diff
@@ -402,6 +402,98 @@ paths:
...
+      responses:
+        '201':
+          description: Software entry created successfully
+        '400':
+          description: Bad request, software entry already exists or invalid data
```
Jdallos: It looks like specs don't typically define specific error responses, see https://github.com/Metatavu/time-bank-api-spec/blob/develop/swagger.yaml for some examples.
Rather they seem to use "default".
Here is another good and larger example if needed:
https://github.com/Metatavu/etra-cloud-api-spec/blob/24e30f958a099d9681ebd3193c06d04c95c0bdd7/swagger.yaml

You should update these for all endpoints

swagger.yaml
Jdallos: Update all endpoints with operationId and tags

swagger.yaml
```diff
@@ -639,4 +731,44 @@ components:
...
+          type: string
+          format: uri
+          description: URL to relevant information/marketing page
+        image:
+          type: string
+          format: uri
+          description: URL to the image of the software
+        status:
+          type: string
+          enum: [pending, under review, accepted, deprecated, declined]
```
Jdallos: It is probably worth creating some kind of enum schema for these

Jdallos (changes requested):

swagger.yaml
```diff
@@ -402,6 +402,115 @@ paths:
...
+      requestBody:
+        description: Software entry to register
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/SoftwareRegistry"
+      responses:
+        "201":
+          description: Software entry created successfully
```
Jdallos, line 437: Is there no content returned as part of this response?

swagger.yaml
```diff
@@ -639,4 +748,47 @@ components:
...
         paid:
           type: boolean
-          description: Status of on call week payment
\ No newline at end of file
+          description: Status of on call week payment
+
+    SoftwareStatus:
+      type: string
+      enum: [pending, under review, accepted, deprecated, declined]
+      description: Status of the software entry
```
Jdallos: Here is an example of how an enum schema has been defined in the time-bank, I would recommend doing this in a similar way:

    Timespan:
      type: string
      description: Enum for timespan
      enum:
        - ALL_TIME
        - YEAR
        - MONTH
        - WEEK

Jdallos (approved):
Approved with a couple of comments

PR #20, opened 2024-09-03
Questionnaire spec design

### LINKED ISSUE #19: Questionnaire spec design

McDuuck:


swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
         "default":
           description: Software entry not found
 
+  "/questionnaire":
+    get:
+      operationId: listQuestionnaire
```
Iiro127: plural

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Questionnaire"
+        "default":
+          description: Bad request, questionnaire entry already exists or invalid data
+
+  "/questionnaire/{id}":
+    get:
+      operationId: getQuestionnaireById
```
Iiro127, line 635: No need to specify "by id". getQuestionnaire is enough, or perhaps findQuestionnaire?

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+          description: Questionnaire entry updated successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Questionnaire"
+        "default":
+          description: Bad request, invalid data
+
+    delete:
+      operationId: deleteQuestionnaireById
```
Iiro127: See earlier comment

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+          description: Questionnaire entry retrieved successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Questionnaire"
+        "default":
+          description: Questionnaire entry not found
+
+    put:
+      operationId: updateQuestionnaireById
```
Iiro127: See earlier comment

Iiro127 (changes requested):

Iiro127 (approved):
nice

swagger.yaml
AfoninOleksandr: ++

swagger.yaml
AfoninOleksandr: sure

AfoninOleksandr (approved):

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
         "default":
           description: Software entry not found
 
+  "/questionnaire":
+    get:
+      operationId: listQuestionnaires
+      summary: Get all questionnaire entries
```
```suggestion
      summary: List questionnaires
```

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+      description: Retrieve a list of all questionnaire entries in the registry.
```
```suggestion
      description: Retrieve a list of all questionnaires in the registry.
```

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+      tags:
+        - Questionnaire
+      responses:
+        "200":
+          description: A list of questionnaire entries
```
```suggestion
          description: A list of questionnaires
```

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+              $ref: "#/components/schemas/Questionnaire"
+      responses:
+        "201":
+          description: Questionnaire entry created successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Questionnaire"
+        "default":
+          description: Bad request, questionnaire entry already exists or invalid data
```
Jdallos: I think this "default" response and all the other endpoints (including the list which does not currently have the default response) could be improved by using something like this:
"default":
          content:
            application/json:
              schema:
                type: array
                items:
                  
It's taken from the time-bank-v2-spec and would require adding an error schema

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Questionnaire"
+
+    post:
+      operationId: createQuestionnaire
+      summary: Register new questionnaire
+      description: Register a new questionnaire entry in the registry.
```
Jdallos: I think we can just refer to Questionnaire or questionnaires rather than questionnaire entry. Please remove all references to entry/ entries- if you agree.

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Questionnaire"
+        "default":
+          description: Bad request, questionnaire entry already exists or invalid data
+
+  "/questionnaire/{id}":
+    get:
+      operationId: getQuestionnaire
```
```suggestion
      operationId: getQuestionnaireById
```

swagger.yaml
```diff
@@ -532,6 +532,119 @@ paths:
...
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: ID of the questionnaire entry to delete
+          schema:
+            type: string
+      responses:
+        "204":
+          description: Questionnaire entry deleted successfully
```
```suggestion
          description: Empty response indicating successful removal.
```

swagger.yaml
```diff
@@ -844,4 +957,37 @@ components:
...
+          description:
+            type: string
+            description: Description of the questionnaire
+          questions:
+            type: array
+            items:
+              $ref: '#/components/schemas/QuestionnaireQuestions'
+            description: List of questions in the questionnaire
+
+      QuestionnaireQuestions:
```
Jdallos: Could this just be Questions?

swagger.yaml
```diff
@@ -844,4 +957,37 @@ components:
...
+        type: object
+        properties:
+          question:
+            type: string
+            description: The question text
+          answers:
```
Jdallos: Could this be options rather than answers?

swagger.yaml
```diff
@@ -844,4 +957,37 @@ components:
         lastUpdatedAt:
           type: string
           format: date-time
-          description: Date and time when the software entry was last updated
\ No newline at end of file
+          description: Date and time when the software entry was last updated
+        
+      Questionnaire:
```
Jdallos: Few thoughts about potential missing properties:
-Do we need some kind of passingScore attribute? 
-How will we record who has passed a questionnaire? Perhaps it makes more sense to record the questionnaires passed on the user entity Iiro is working on but what do you think? We might also want to record the score for a user, rather than just if they passed or failed.
-It would be very nice if the Questionnaire could also have some kind of tags for example this would allow us to show if it id targeted to a developer or a designer, or perhaps if its related to security vs company policy etc.

Jdallos (changes requested):

swagger.yaml
```diff
@@ -540,6 +685,23 @@ components:
...
+        - message
+      properties:
+        code:
+          type: integer
+          format: int32
+          description: Error code
+        message:
+          type: string
+          description: Error message
+    # This is from time-bank-api-spec swagger.yaml file
```
Jdallos: No need for this comment

```suggestion
```

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+        title:
+          type: string
+          description: Title of the questionnaire
+        description:
+          type: string
+          description: Description of the questionnaire
+        options:
+          type: array
+          items:
+            $ref: '#/components/schemas/Questions'
```
Jdallos: Use double quotes

swagger.yaml
```diff
@@ -532,6 +532,151 @@ paths:
...
+                items:
+                  $ref: "#/components/schemas/Questionnaire"
+      "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Error"
```
Jdallos: This needs to be indented by another tab. Remember to past the spec into https://editor.swagger.io/ which will highlight these kinds of errors

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+          type: array
+          items:
+            type: string
+          description: List of tags for the questionnaire
+        passedUsers:
+          type: array
+          items:
+            type: integer
+          description: List of users who have passed the questionnaire
+        passedScore:
```
```suggestion
        passScore:
```

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+    Questions:
+      type: object
+      properties:
+        question:
+          type: string
+          description: The question text
+        options:
+          type: array
+          items:
+            type: boolean
```
Jdallos: If this is a list of possible answers then why is it an array of booleans?

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+          description: List of possible answers
+        correctAnswers:
+
```
Jdallos, line 1138:
```suggestion
```

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+          type: array
+          items:
+            type: boolean
```
Jdallos: Again I'm not sure why this is a list of booleans perhaps we should discuss?

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+        passedUsers:
+          type: array
+          items:
+            type: integer
+          description: List of users who have passed the questionnaire
+        passedScore:
+          type: integer
+          description: Minimum score required to pass the questionnaire
+
+    Questions:
```
Jdallos, line 1127: It would be nice if the questionnaire to begin with can handle both single select and mutli select. Does the current schema allow this?

swagger.yaml
```diff
@@ -844,4 +1006,53 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/Questions'
+          description: List of questions in the questionnaire
+        tags:
+          type: array
+          items:
+            type: string
+          description: List of tags for the questionnaire
+        passedUsers:
```
Jdallos, line 1118: I wonder if we can come up with a more descriptive name, I feel passedUsers is a bit vague. But I can't think of a better name so it can be left as is if you think this is the best option.

Jdallos (changes requested):

swagger.yaml
```diff
@@ -844,4 +1005,54 @@ components:
...
+          type: array
+          items:
+            type: object
+            properties:
+              label:
+                type: string
+                description: Display text for the option
+              value:
+                type: boolean
+                description: Boolean value of the option
```
```suggestion
                description: Boolean value to indicate if option is correct or incorrect
```

swagger.yaml
```diff
@@ -844,4 +1005,54 @@ components:
...
+          description: List of correct answers with labels and boolean values
```
Jdallos: From what I understand this description is not correct- it's a list of answers,which includes the correct answer, not a list of correct answers.

Something like:
List of options objects containing options labels and answer status boolean

swagger.yaml
```diff
@@ -844,4 +1005,54 @@ components:
...
+          type: integer
+          description: Minimum score required to pass the questionnaire
+
+    Questions:
+      type: object
+      properties:
+        question:
+          type: string
+          description: The question text
+        options:
```
Jdallos, line 1133: Perhaps options should also have it's own schema?

Jdallos (approved):

PR #22, opened 2024-09-03
User schema and endpoints

### LINKED ISSUE #21: User endpoints and schema

Iiro127:


swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+      required:
+        - id
+        - firstName
+        - lastName
+        - isActive
+      properties:
+        id:
+          type: string
+          format: uuid
+          description: Person ID in Keycloak.
```
McDuuck: Instead of "Person" it should be "Users"

swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+        firstName:
+          type: string
+          description: Person first name.
```
McDuuck: See above comment

swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+        lastName:
+          type: string
+          description: Person last name.
```
McDuuck: See above comment

swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+        email:
+          type: string
+          description: Person email.
```
McDuuck: See above comment

swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+        isActive:
+          type: boolean
+          description: Defining is person active e.g. is employed.
```
McDuuck: See above comment

swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+        severaGuid:
+          type: string
+          description: Persons GUID in Severa
```
McDuuck: See above comment

swagger.yaml
```diff
@@ -540,6 +592,39 @@ components:
...
+        forecastId:
+          type: integer
+          format: int32
+          description: Persons ID in Forecast. Used for keeping a reference of Forecast based TimeEntries and Users.
```
McDuuck: See above comment

McDuuck (changes requested):

McDuuck (approved):
Good work mate!

dalju005-xamk (approved):
Nice work, Dude.

swagger.yaml
```diff
@@ -532,6 +532,58 @@ paths:
...
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/User"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
```
Jdallos: Error schema does not currently exists, it's good to use it like this but it needs to be created first.

Jdallos: Also pasting your spec in the swagger editor would highlight issues like this

swagger.yaml
```diff
@@ -532,6 +532,58 @@ paths:
         "default":
           description: Software entry not found
 
+  /users:
```
Jdallos: Lets be consistent to the style of this spec, endpoints are wrapped in double quotes

```suggestion
  "/users":
```

swagger.yaml
```diff
@@ -532,6 +532,58 @@ paths:
...
+                items:
+                  $ref: "#/components/schemas/User"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  /users/{userId}:
```
```suggestion
  "/users/{userId}":
```

swagger.yaml
```diff
@@ -532,6 +532,58 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/User"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
```
Jdallos: Error schema is missing

swagger.yaml
```diff
@@ -540,6 +592,43 @@ components:
...
+        email:
+          type: string
+          description: User's email.
+        isActive:
+          type: boolean
+          description: Defining is user is active e.g. is employed.
+        passedQuestionnaires:
+          type: array
+          items:
+            type: integer
```
Jdallos, line 615: An array of integers? Can you explain how this works, I would maybe have expected an array of strings/ questionnaire Ids or perhaps an array of objects with the Q id and the user score?

swagger.yaml
```diff
@@ -540,6 +592,43 @@ components:
...
+        lastName:
+          type: string
+          description: User's last name.
+        email:
+          type: string
+          description: User's email.
+        isActive:
+          type: boolean
+          description: Defining is user is active e.g. is employed.
+        passedQuestionnaires:
```
Jdallos, line 612: I wonder if completedQuestionnaires might be a better name- what do you think?

Jdallos (changes requested):

swagger.yaml
Iiro127: The integer represents the questionnaire id the user has passed, as the id is in integer format.

swagger.yaml
Iiro127: Not sure about that. A questionnaire can be completed and not passed. I don't think we need to know what questionnaires have been failed?

swagger.yaml
Jdallos: Is there a good reason to have the id as integer rather than uuid string? I'm not sure- up to you

swagger.yaml
Jdallos: I thought it was a good idea to add the Error schema and reference it as the default response as it was previously. I don't see any reason for the schema to not just be the same as in time bank

Jdallos (approved):

PR #24, opened 2024-09-03
updated the software registry endpoints

### LINKED ISSUE #23: updated the software registry endpoints

yuusufisse:


swagger.yaml
```diff
@@ -805,11 +805,17 @@ components:
...
+        - DEPRECATED
         - DECLINED
 
     SoftwareRegistry:
       type: object
+      required:
+        - name
+        - description
+        - url
+        - image
```
Jdallos, line 817: Are you sure image should be required? Probably most things have an image but it seems to create a bit more work for the users- unless you are able to get the image straight from the Url (which would be very cool).

swagger.yaml
```diff
@@ -844,4 +850,16 @@ components:
         lastUpdatedAt:
           type: string
           format: date-time
-          description: Date and time when the software entry was last updated
\ No newline at end of file
+          description: Date and time when the software entry was last updated
+        recommend:
```
Jdallos, line 854: Needs description

swagger.yaml
```diff
@@ -844,4 +850,16 @@ components:
...
+          type: array
+          items:
+            type: string
+        tags:
```
Jdallos, line 859: Needs description

swagger.yaml
```diff
@@ -844,4 +850,16 @@ components:
...
+          type: array
+          items:
+            type: string
+        users:
```
Jdallos, line 864: Needs description

Jdallos (changes requested):

swagger.yaml
yuusufisse: image is required for the logo of the software

Jdallos (approved):

PR #28, opened 2024-10-02
Adding flextime path from SeveraAPI

### LINKED ISSUE #27: Balance restoration SeveraAPI specs
Starting to change Balance for MetaHome to use SeveraAPI
Include relevant lambas from SeveraAPI

dalju005-xamk:
Added flextime path.
Security things and FlextimeModel based on Severa Public Rest API Documentation

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -760,13 +760,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/FlextimeModel'
+                $ref: "#/components/schemas/FlextimeModel"
         "403":
           description: Addon required
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Error'
+                $ref: "#/components/schemas/Error"
         "404":
           description: Resource not found
         default:
@@ -774,7 +774,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Error'
+                $ref: "#/components/schemas/Error"
 
 components:
   securitySchemes:
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -4,8 +4,14 @@ info:
   description: Metatavu Home lambdas
   version: 1.0.0
 
+servers:
+  - url: https://api.severa.visma.com/rest-api
+
 security:
   - bearerAuth: []
+  - OAuth2:
+    - users:read
+  - ClientIdAuth: []
 
 paths:
   "/allocations":
@@ -721,12 +727,77 @@ paths:
                 items:
                   $ref: "#/components/schemas/Error"
 
+  "/v1/users/{userGuid}/flextime": 
+    get:
+      tags:
+        - UsersRead
+      summary: 
+        Get the flextime balance for a user for a specified date. Total balance
+        is returned for the given date. Month balance is the balance for the
+        month of the given date. Values are returned only if the advanced time
+        tracking add-on is active.
+      operationId: Flextime_GetFlextime
+      parameters:
+        - name: userGuid
+          in: path
+          required: true
+          description: Id of the user.
+          schema:
+            type: string
+          x-position: 1
+          x-relation-guid: UserOutputModel
+        - name: eventDate
+          in: query
+          description:
+            Date for which to get the balance. Max 12 months into the future.
+          schema:
+            type: string
+            format: date-time
+          x-position: 2
+      responses:
+        "200":
+          description: FlextimeModel.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FlextimeModel'
+        "403":
+          description: Addon required
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        "404":
+          description: Resource not found
+        default:
+          description: Default error response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
 components:
   securitySchemes:
     bearerAuth:
       type: http
       scheme: bearer
       bearerFormat: JWT
+
+    OAuth2:
+      type: oauth2
+      description: SeveraAPI uses OAuth 2 with the Client Credentials flow
+      flows:
+        clientCredentials:
+          tokenUrl: /token
+          refreshUrl: /refreshtoken
+          scopes:
+            users:read: Read-only access to user resources
+
+    clientIdAuth:
+      type: apiKey
+      description: Client_Id header
+      name: Client_Id
+      in: header
 
   schemas:
     Error:
@@ -1167,3 +1238,15 @@ components:
         value:
           type: boolean
           description: Boolean value to indicate if option is correct or incorrect
+
+    FlextimeModel:
+      type: object
+      properties:
+        totalFlextimeBalance:
+          type: number
+          format: double
+          nullable: true
+        monthFlextimeBalance:
+          type: number
+          format: double
+          nullable: true
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -739,7 +739,7 @@ paths:
             type: string
       responses:
         "200":
-          description: FlextimeModel.
+          description: Flextime entry retrieved successfully
           content:
             application/json:
               schema:
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -721,7 +721,7 @@ paths:
                 items:
                   $ref: "#/components/schemas/Error"
 
-  "/severa/flextime/{severaGuid}": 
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+            type: string
+            format: date-time
+          x-position: 2
+      responses:
+        "200":
+          description: FlextimeModel.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FlextimeModel'
```
McDuuck: double quotes

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/Error'
+        "404":
+          description: Resource not found
+        default:
+          description: Default error response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
McDuuck: double quotes

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FlextimeModel'
+        "403":
+          description: Addon required
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
McDuuck: double quotes

swagger.yaml
```diff
@@ -4,8 +4,14 @@ info:
...
   version: 1.0.0
 
+servers:
+  - url: https://api.severa.visma.com/rest-api
+
 security:
   - bearerAuth: []
+  - OAuth2:
+    - users:read
+  - ClientIdAuth: []
```
Jdallos: This should not be necessary. Remember the specification is describing the lambda function we are calling, not whatever api's the lambda function is calling. We already have bearer auth set up for accessing the lambdas and this is sufficient.

swagger.yaml
```diff
@@ -4,8 +4,14 @@ info:
   description: Metatavu Home lambdas
   version: 1.0.0
 
+servers:
+  - url: https://api.severa.visma.com/rest-api
+
```
Jdallos: I've not seen this used before in any of our specs and am quite sure it's not necessary.

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
                 items:
                   $ref: "#/components/schemas/Error"
 
+  "/v1/users/{userGuid}/flextime":
```
Jdallos: This looks like the severa endpoint. But we don't need to define that in this spec. Rather I would think this could be something like severa/flexitime/{severaGuid}. 

We should also consider at this point is it possible/ necessary to have an option to not pass an id and get all users flexitime. This could be a nice admin feature but depends on the severa API. If the severa flexitime endpoint always requires an id and does not return a list then we can not worry about this for now and just implement it with the endpoint as I just described above.

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+          refreshUrl: /refreshtoken
+          scopes:
+            users:read: Read-only access to user resources
+
+    clientIdAuth:
+      type: apiKey
+      description: Client_Id header
+      name: Client_Id
+      in: header
+
```
Jdallos: I think this can be removed

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/Error"
+        "404":
+          description: Resource not found
+        default:
+          description: Default error response
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
```
Jdallos: It look like this is copied from the severa flexitime end point but this is not necessary. In our case we just want to provide the balance for the current day (and potentially month) in both cases this means no eventDate param is required,
Lets refer to userGuid as severaGuid as this is what is is called in out user schema.

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+  "/v1/users/{userGuid}/flextime": 
+    get:
+      tags:
+        - UsersRead
+      summary: 
+        Get the flextime balance for a user for a specified date. Total balance
+        is returned for the given date. Month balance is the balance for the
+        month of the given date. Values are returned only if the advanced time
+        tracking add-on is active.
+      operationId: Flextime_GetFlextime
```
Jdallos: use camelCase and similar naming conventions as for all other endpoints in this schema e.g. getUserFlexitime

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+          x-position: 1
+          x-relation-guid: UserOutputModel
+        - name: eventDate
+          in: query
+          description:
+            Date for which to get the balance. Max 12 months into the future.
+          schema:
+            type: string
+            format: date-time
+          x-position: 2
```
```suggestion
```

swagger.yaml
```diff
@@ -721,13 +727,78 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/FlextimeModel"
+        "403":
+          description: Addon required
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "404":
+          description: Resource not found
```
```suggestion
```

swagger.yaml
```suggestion
```

swagger.yaml
```diff
@@ -1167,3 +1238,15 @@ components:
         value:
           type: boolean
           description: Boolean value to indicate if option is correct or incorrect
+
+    FlextimeModel:
```
```suggestion
    Flextime:
```

swagger.yaml
```diff
@@ -1167,3 +1238,15 @@ components:
...
+      type: object
+      properties:
```
Jdallos, line 1203: Consider whether we should also return the userSeveraId as part of this schema- it's probably not necessary but still worth considering. Rememeber we are not bound to have this be identical to what the serveraAPI returns.

Jdallos (changes requested):

swagger.yaml
dalju005-xamk: I think this Severa endpoint just returns flextime from one user. It needs to have GUID and optional date. 
I change userGuid to our find-user spec severaGuid.

swagger.yaml
dalju005-xamk: This is from Severa Docs: Endpoint get the flextime balance for a user for a specified date. Total balance is returned for the given date. Month balance is the balance for the month of the given date.
Im not sure if date is not given, will it use current date as default.

swagger.yaml
dalju005-xamk: Would this be used when Admin wants to list employees flextime. Like there would return severaId and balances ?

swagger.yaml
```diff
@@ -721,6 +721,30 @@ paths:
                 items:
                   $ref: "#/components/schemas/Error"
 
+  "/severa/flextime/{severaGuid}": 
+    get:
+      tags:
+        - UsersRead
```
```suggestion
        - Flextime
```

swagger.yaml
```diff
@@ -721,6 +721,30 @@ paths:
...
+      summary: 
+        Get the flextime balance for a user. Total balance
+        is returned for the current date. Month balance is the balance for the
+        month of the current date.
+      operationId: getFlextimeByUser
```
Jdallos: Looking at the ordering of the other endpoints in the spec its typically:
operationId
summary
description
tags
parameters
responses

Please reorder accordingly

```suggestion
      operationId: getFlextimeBySeveraGuid
```

swagger.yaml
```diff
@@ -721,6 +721,30 @@ paths:
...
                   $ref: "#/components/schemas/Error"
 
+  "/severa/flextime/{severaGuid}": 
+    get:
+      tags:
+        - UsersRead
+      summary: 
+        Get the flextime balance for a user. Total balance
+        is returned for the current date. Month balance is the balance for the
+        month of the current date.
```
```suggestion
        Get the flextime balance for a user.
```

swagger.yaml
```diff
@@ -721,6 +721,30 @@ paths:
...
+          description: severaGuid of the user.
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Flextime entry retrieved successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Flextime"
```
Jdallos, line 744: There should also be a default error response as was done for questionnaries:
 "default":
            content:
              application/json:
                schema:
                  type: array
                  items:

Jdallos (changes requested):

tonipulkkinenn (approved):

PR #31, opened 2024-10-17
Feature 29 vacation spec design

naStipe:
Implemented vacation request  specs design according to time-bank-API-spec

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -917,6 +917,182 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  "/vacationRequests/{id}/statuses":
+    get:
+      operationId: listVacationRequestStatuses
+      summary: List vacation requests statuses
+      description: List vacation requests statuses
+      tags:
+        - VacationRequestStatus
+      parameters:
+        - name: id
+          required: true
+          in: path
+          description: VacationRequest id
+          schema:
+            type: string
+            format: uuid
+      responses:
+        '200':
+          description: List of vacation requests.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/VacationRequestStatus'
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Error"
+    post:
+      operationId: createVacationRequestStatus
+      summary: Post a new vacation request status
+      description: Add new vacation request status
+      tags:
+        - VacationRequestStatus
+      parameters:
+        - name: id
+          required: true
+          in: path
+          description: VacationRequest id
+          schema:
+            type: string
+            format: uuid
+      requestBody:
+        description: New vacation request status
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/VacationRequestStatus'
+      responses:
+        '201':
+          description: created new vacation request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VacationRequestStatus'
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -1198,6 +1540,119 @@ components:
...
+          description: Vacation days used for requested vacation
+        type:
+          $ref: "#/components/schemas/VacationType"
+        message:
+          type: string
+          description: Message attached to request
+        createdBy:
+          type: string
+          format: uuid
+          description: ID of person who created status in UUID
```
dalju005-xamk, line 1603: Person's Keycloak ID ?

swagger.yaml
```diff
@@ -1198,6 +1540,119 @@ components:
           type: boolean
           description: Boolean value to indicate if option is correct or incorrect
 
+    VacationType:
+      type: string
+      description: Enum for vacation type
+      enum:
+        - VACATION
```
dalju005-xamk, line 1547: These should be inline with company policies and collective labor agreements.

swagger.yaml
```diff
@@ -750,7 +750,349 @@ paths:
...
+      responses:
+        "204":
+          description: Empty response indicating successful removal.
+        "default":
+          description: Invalid request was sent to the server
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+  "/vacationRequests/{id}/statuses":
```
dalju005-xamk, line 920: Not sure are these vacationRequest status thing so needed. Could these be handled some other way ?

dalju005-xamk (approved):
Nice work dudes. Not sure if all this is needed, but that is something to find out in future.

swagger.yaml
Hanhz2707: yeah, createdBy: / type: string / format: uuid / description: Person keycloak ID in UUID

AfoninOleksandr (approved):
Looks good

swagger.yaml
```diff
@@ -750,7 +750,349 @@ paths:
...
+            type: boolean
+      responses:
+        '200':
+          description: List of vacation requests.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/VacationRequest'
```
Jdallos: Lets update all usages of single quotes with double quotes.

swagger.yaml
Jdallos: In the original design it was decided that the vacation status should be a separate entity, this allows us to record multiple vacation "responses" e.g. approval or decline along with accompanying messages/ notes.

Feel free to have a think if you want to change this and we could have further discussion. At least IMO vacationRequestStatuses are fine for now as the UI is already build around them being a separate entity.

swagger.yaml
Jdallos: Yes these might need to be updated- I'd suggest sending Taneli or Eetu a message about this current list and asking them what other options are used e.g. in Severa.

swagger.yaml
```diff
@@ -1198,6 +1540,119 @@ components:
...
+      required:
+        - personId
+        - days
+        - startDate
+        - endDate
+        - type
+        - message
+        - createdAt
+        - updatedAt
+        - draft
```
Jdallos: Maybe this does not need to be a required parameter?

swagger.yaml
Jdallos: This id will be coming from the users schema. The user id is the same as keycloak Id.

Jdallos (approved):
Approved with some suggestions

swagger.yaml
naStipe: Then I guess we will keep them for now

swagger.yaml
Hanhz2707: updated

swagger.yaml
Hanhz2707: deleted

PR #35, opened 2024-10-31
feature-34-home-lambda-api-specs-for-sprint-view

### LINKED ISSUE #34: Home Lambda API specs for sprint-view
Creating lambda API specs for functions to retrieve resourceAllocations, Projects, Phases(Tasks before), and WorkHours

naStipe:
Added schemas and function specs for ResourceAllocations, WorkHours and Phases

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -789,7 +789,7 @@ paths:
       tags:
         - resourceallocations
       parameters:
-        - name: severaGuid
+        - name: severaUserGuid
           in: path
           required: true
           description: severaGuid of the user.
@@ -821,7 +821,7 @@ paths:
       tags:
         - WorkHours
       parameters:
-        - name: severaWorkHourGuid
+        - name: severaUserGuid
           in: path
           required: true
           description: ID of the work hour.
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -727,7 +727,7 @@ paths:
       summary: Get the flextime balance for a user.
       description: Retrieve the flextime balances for a user based on their Severa GUID.
       tags:
-        - flextime
+        - FlexTime
       parameters:
         - name: severaUserGuid
           in: path
@@ -787,7 +787,7 @@ paths:
       summary: Get the resource allocations for a user.
       description: Retrieve the resource allocations for a user based on their Severa GUID.
       tags:
-        - resourceallocations
+        - ResourceAllocations
       parameters:
         - name: severaUserGuid
           in: path
@@ -1787,9 +1787,9 @@ components:
           type: number
           format: double
         phase:
-           $ref: "#/components/schemas/BaseSubModel"
+           $ref: "#/components/schemas/PhaseSubModel"
         user:
-          $ref: "#/components/schemas/BaseSubModel"
+          $ref: "#/components/schemas/UserSubModel"
         project:
           $ref: "#/components/schemas/ResourceAllocationProjectSubModel"
 
@@ -1799,11 +1799,11 @@ components:
         severaWorkHoursGuid:
           type: string
         user:
-          $ref: "#/components/schemas/BaseSubModel"
+          $ref: "#/components/schemas/UserSubModel"
         project:
           $ref: "#/components/schemas/PhaseProjectSubModel"
         phase:
-          $ref: "#/components/schemas/BaseSubModel"
+          $ref: "#/components/schemas/PhaseSubModel"
         description:
           type: string
         eventDate:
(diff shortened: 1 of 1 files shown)


Hanhz2707 (approved):

AfoninOleksandr (approved):
Looks good

swagger.yaml
```diff
@@ -1436,21 +1528,18 @@ components:
           description: Unique identifier for the software entry
         name:
           type: string
-          description: Name of the software entry
```
dalju005-xamk, line 1439: What are these?

swagger.yaml
```diff
@@ -721,15 +721,15 @@ paths:
                 items:
                   $ref: "#/components/schemas/Error"
 
-  "/severa/{severaGuid}/flextime":
+  "/severa/{severaUserGuid}/flextime":
```
dalju005-xamk: This needs to be changes to other places also. Where it now severaGuid. In example in keycloak-Api-service. Maybe it on bugfix PR ?

dalju005-xamk (approved):

swagger.yaml
Jdallos: I thought we discussed the idea of renaming all references to a Guid to Id. I think this is the best options. So in this case it should be severaUserId. Please ensure all other references to Guid are upated

Jdallos: I won't leave any other comments relating to this change so please check

swagger.yaml
```diff
@@ -751,7 +751,99 @@ paths:
                 items:
                   $ref: "#/components/schemas/Error"
 
-  "/vacationsRequests":
+  "/severa/projects/{severaProjectGuid}/phaseswithhierarchy":
```
Jdallos: I'm not sure about the endpoint name of phaseswithhierarchy, I'd suggest just phases

swagger.yaml
```diff
@@ -751,7 +751,99 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Error"
+
+  "/severa/users/{severaUserGuid}/workhours":
+    get:
+      operationId: getWorkHoursBySeveraProjectId
```
Jdallos: Is it the workhours for a project?  I see only the userId in use, how do we define which project? Or is the id and summary incorrectly named?

swagger.yaml
```diff
@@ -751,7 +751,99 @@ paths:
...
+      summary: Get all the work hours for project
+      description: Retrieve all the work hours for a particular user with given ID.
+      tags:
+        - WorkHours
+      parameters:
+        - name: severaUserGuid
+          in: path
+          required: true
+          description: ID of the work hour.
```
Jdallos: Isn't it the ID of the user?

swagger.yaml
Jdallos: Looking at the workhours schema I assume we can define the project/ phase the workhours refer to. I assume this is done using query params, in which case there query params should be defined here. Are some of them optional?

swagger.yaml
```diff
@@ -751,7 +751,99 @@ paths:
...
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Error"
+
+  "/severa/{severaUserGuid}/resourceallocations":
```
Jdallos: Use camelCase for the endpoint naming e.g. resourceAllocations

Jdallos (changes requested):

swagger.yaml
naStipe: There is no endpoint "phases" for user. this endpoint receives all phases for a particular user. Do you want us to retrieve all phases and then sort the ones that we need?

swagger.yaml
naStipe: Yes. All of them are optional, but we will add the lacking params

swagger.yaml
naStipe: Yes, it is. We will fix the description.

swagger.yaml
naStipe: Does it matter how they are written in severa? These endpoint were "copied" from the docs

swagger.yaml
naStipe: Actually, they are not defined in the params. Project/phase is a property of workhours. So they should be stored in the same entity in the database, not defined manually in params. Let me know if I misunderstood your question.

swagger.yaml
naStipe: All phases of a particular project* 
Not user

swagger.yaml
```diff
@@ -751,7 +751,132 @@ paths:
...
+          schema: 
+            type: string
+            format: date-time
+        - name: eventDate
+          required: false
+          in: query
+          description: Filter work hours by eventDate
+          schema: 
+            type: string
+            format: date
```
naStipe, line 862: Not sure if we need to filter by eventDate. Maybe we should delete it for now?

swagger.yaml
```diff
@@ -751,7 +751,125 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Phase"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
```
```suggestion
```

Jdallos: I noticed this crept into the spec at some point, we don't need these two lines so they can be removed.

swagger.yaml
```diff
@@ -751,7 +751,125 @@ paths:
...
+                  $ref: "#/components/schemas/ResourceAllocations"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
```
```suggestion
```

Jdallos (approved):

PR #37, opened 2024-11-03
feature-36-home-lambda-api-spec-for-memo-management

DZotoff:


swagger.yaml
```diff
@@ -1093,6 +1094,264 @@ paths:
...
+                  cardId:
+                    type: string
+                    description: The ID of the Trello card.
+                  comment:
+                    type: string
+                    description: The comment text.
+        "400":
+          description: Missing required parameters.
+        "500":
+          description: Failed to create comment.
```
Jdallos: I think the best option is to have a generic "default" error response like we do for other end points rather than specifying particular error statuses. Here is what we are using already:

      "default":
          content:
            application/json:
              schema:
                  
I'd suggest replacing all 400 and 500 responses like you have done here with this.

swagger.yaml
```diff
@@ -1093,6 +1094,264 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/card-comments":
```
Jdallos: Id suggest having it as trello/card-comments. Similar for google drive related endpoints also.

Jdallos: Let me know what you think, if you agree this should be applied to all relevant endpoints

swagger.yaml
```diff
@@ -1093,6 +1094,264 @@ paths:
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  cardId:
+                    type: string
+                  title:
+                    type: string
+                  description:
+                    type: string
```
Jdallos: Probably worth making a TrelloCard schema for this

swagger.yaml
```diff
@@ -1093,6 +1094,264 @@ paths:
...
+                  comment:
+                    type: string
+                    description: The comment text.
+        "400":
+          description: Missing required parameters.
+        "500":
+          description: Failed to create comment.
+
+  "/card":
+    post:
```
Jdallos, line 1252: Throughout the spec we typically order the get (list) and then post (create) for a given endpoint. Can you make sure where relevant you order your new endpoints in the same way- it just makes it easier to check the spec as it grows.

swagger.yaml
Jdallos: Actually it seems you already have created the schema Card. So it should be referenced here also

swagger.yaml
```diff
@@ -1093,6 +1094,264 @@ paths:
...
+      responses:
+        "200":
+          description: List of cards retrieved successfully.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Card"
+
```
```suggestion
```

swagger.yaml
```diff
@@ -1093,6 +1094,264 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Member"
+        "500":
+          description: Failed to fetch board members' emails.
+
+  "/card/{id}":
```
Jdallos: Typically we would want the card endpoints to be grouped together, so this card/{id} should come straight after / card

swagger.yaml
```diff
@@ -1661,4 +1920,107 @@ components:
...
+          items:
+            type: string
+          description: List of attachments related to the memo
+        assignedTo:
+          type: array
+          items:
+            type: string
+          description: List of users assigned to this memo
+
+    Card:
```
Jdallos: Consider whether these schemas should be named more specifically e.g. googleDriveMemo, TrelloCard, TrelloMember. It's your call

Jdallos (changes requested):

swagger.yaml
DZotoff: Return type of functions is a single object, why is an array used here?

Jdallos: Well spotted, I've updated the code snippet so it's now correct. Not sure how or when that crept in, feel free to update by removing the
type: array
items:
lines from other default error responses in the spec.

Jdallos (approved):

PR #48, opened 2024-11-13
changes to vacationrequest related schemas

### LINKED ISSUE #43: Modifying vacation request status.
Removing some of vacation request status properties ann removing the endpoints for it

rimz178:


swagger.yaml
```diff
@@ -335,9 +335,8 @@ paths:
       tags:
         - Interest
       parameters:
-        - name: dealId
-          in: path
-          required: true
+        - name: dealIdgit fetch origin
+git checkout feature-43-modifying-vacation-request-status
```
dalju005-xamk: What is this ?

swagger.yaml
```diff
@@ -857,10 +856,10 @@ paths:
       tags:
         - VacationRequests
       parameters:
-        - name: personId
+        - name: userId
           required: false
```
dalju005-xamk, line 861: Does this false need to be here ? Can it be removed if nothing is required ?

dalju005-xamk (changes requested):

swagger.yaml
naStipe: Or actually maybe it should be required...

swagger.yaml
```diff
@@ -335,9 +335,7 @@ paths:
       tags:
         - Interest
       parameters:
-        - name: dealId
-          in: path
-          required: true
+        - name: dealId
```
naStipe: I suppose that those were not supposed to be deleted

swagger.yaml
```diff
@@ -1648,7 +1474,7 @@ components:
       type: object
       description: Person's single vacation request
       required:
-        - personId
+        - userId
         - days
         - startDateS
```
naStipe: this renamed into startDate

swagger.yaml
```diff
@@ -1705,25 +1533,12 @@ components:
...
-          type: string
-          format: uuid
-          description: ID of vacation request status in UUID
-          default: null
-        vacationRequestId:
-          type: string
-          format: uuid
-          description: ID of vacation request in UUID
-          default: null
+
```
naStipe: empty line is not necessary here

swagger.yaml
```diff
@@ -1732,11 +1547,6 @@ components:
           type: string
           format: date-time
           description: Date time of status update
```
naStipe, line 1576: Maybe add an empty line between VacationRequestStatus and Flextime

naStipe (changes requested):

swagger.yaml
naStipe: and also createdBy probably should be required as well

swagger.yaml
```diff
@@ -1683,6 +1509,8 @@ components:
           description: Vacation days used for requested vacation
         type:
           $ref: "#/components/schemas/VacationType"
+        status:
```
naStipe, line 1537: This should be an array. See "options" property for "Questionnaire" for a reference

swagger.yaml
```diff
@@ -334,8 +334,9 @@ paths:
         Updates string where intrests are stored in Pipedrive
       tags:
         - Interest
-      parameters:
-        - name: dealId
+      parameters:in: path
+          required: true
+        - name: dealId
```
naStipe: This part is undelined like the error. Some issues with the structure/indexing probably. Try to fix it

swagger.yaml
```diff
@@ -477,12 +478,7 @@ paths:
         "200":
           description: Software entry retrieved successfully
           content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/SoftwareRegistry"
-        "default":
-          description: Software entry not found
-
+            application/json:and maybe you should make "userId" required for vacationRequests
```
naStipe: Part of the code is missing and there is a comment from GitHub for some reason

naStipe (changes requested):

swagger.yaml
naStipe: it should look like:
type: array
  items:
    $ref:....

naStipe: and the status should be an array, not type

naStipe (changes requested):

naStipe (approved):

Hanhz2707 (approved):

dalju005-xamk (approved):

swagger.yaml
```diff
@@ -857,10 +857,10 @@ paths:
       tags:
         - VacationRequests
       parameters:
-        - name: personId
-          required: false
+        - name: userId
+          required: true
```
Jdallos: We already have the vacations/{id} endpoint to as the find endpoint for vacations. I understand we might want to have userId as a param here if we want to retrieve vacationRequests for specific users- but are we sure it should be required? What is we want to list all vacation requests for all users e.g. in a admin view?

swagger.yaml
```diff
@@ -1703,8 +1531,14 @@ components:
...
           description: Vacation days used for requested vacation
-        type:
+        type: 
           $ref: "#/components/schemas/VacationType"
+          description: type of vacation request
+        status:
+          type: array
+          items:
+            $ref: "#/components/schemas/VacationRequestStatus"
+          description: status of vacation request
```
Jdallos: Fix indentation
```suggestion
            description: status of vacation request
```

Jdallos (changes requested):

swagger.yaml
naStipe: Don`t we need the userID to "connect" vacation requests to specific users? Even if we get all the vacationRequests, i thought, that we need to know who these requests are for. And if there is a vacationRequest without id sent to/received from the server, doesn`t it indicate an issue? Correct me if I misunderstood.

naStipe: Nevermind, it is not needed in that place.

rimz178: Should I change the 'required' field to false?

rimz178: Should I change the 'required' field to false?

naStipe: yes

Jdallos (approved):

PR #54, opened 2024-11-19
bugfix-53 -remove-duplicate

### LINKED ISSUE #53: bugfix remove duplicate property from attributes.

rimz178:


naStipe (approved):

swagger.yaml
```diff
@@ -1518,17 +1517,20 @@ components:
...
           items:
-            type: integer
-        severaUserId:
-          type: string
-          description: User's ID in Severa
-        forecastId:
-          type: integer
-          format: int32
-          description: User's ID in Forecast. Used for keeping a reference of Forecast based TimeEntries and Users.
+            type: string
```
dalju005-xamk, line 1533: Is there reason for change forecastID, I think it could be delete entirely

dalju005-xamk (approved):

swagger.yaml
Jdallos: My understanding of this PR was that it was just going to remove the duplicate severaUserId property from attribute.
So I agree I'm not sure the forecastId type should be changed, but is there actually a reason for us to remove it? Maybe it's fine to remove.

As for adding in the vacationDaysByYear and unspentVacationDaysByYear have these changes been requested by Stepan related to his vacations feature?

naStipe: The ForecastID was left intact, because I am not sure yet if some part of the page will break and stop building without it - that is the only reason. 
As for all the other properties: that is how the Attributes looked in my PR, that was merged. Other merge overwritten the changes somehow and i just told Rikku to revert that to the stage it was two days ago.

naStipe: Actually, now that i check it isn`t there so I guess i had something else in mind a few days ago

naStipe (approved):

swagger.yaml
Jdallos: So do we want vacationDaysByYear and unspentVacationDaysByYear to be included in this PR and the forecastId can be unchanged for now?

naStipe: Yes, I definitely want the vacation days to be back on develop, and we can leave forecast untouched as well.

Jdallos: Alright please revert forecastId to how it was before these changes and then we can approve

Jdallos (changes requested):

swagger.yaml
```diff
@@ -1518,13 +1517,20 @@ components:
...
+            description: Year and unspent vacation days in format "YYYY:DDD".
         passedQuestionnaires:
           type: array
           items:
-            type: integer
-        severaUserId:
-          type: string
-          description: User's ID in Severa
+            type: string
```
Jdallos, line 1533: It looks like you have removed the type:integer from passedQuestionnaires and left the type: string from the now removed severaUserId. In effect changing the type of passedQuestionnaires from integer to string. Was this intentional?

rimz178: I deleted it by accident ,  I will fix that.

naStipe: No need. It was intentional, because it is a list of IDs. Aleksi said that it will be a string, unless im mistaken.

Jdallos: Sounds good

Jdallos (approved):

PR #57, opened 2024-11-21
Feature 34 home lambda api specs for sprint view

Hanhz2707:


naStipe (approved):

dalju005-xamk (approved):

swagger.yaml
```diff
@@ -767,7 +767,7 @@ paths:
               schema:
                   $ref: "#/components/schemas/Error"
 
-  "/severa/{severaUserId}/resourceAllocations":
+  "/severa/users/{severaUserId}/resourceAllocations":
```
Jdallos, line 770: Having userId as a path param rather than an optional query param will mean we always need userId to request the resourceAllocation. DO you think this is fine? I don't think we need to worry too much about admin functionality for now so maybe it is fine, just wanted to mention it.

swagger.yaml
```diff
@@ -2111,7 +2111,19 @@ components:
           type: string
           format: date
         project:
-          $ref: "#/components/schemas/PhaseProjectSubModel"
+          type: object
+          description: A reference to Project Sub Module in Phase Object
```
Jdallos: We should remove any mention of sub model/ sub module in descriptions if we are no longer having submodel schemas

Jdallos: This applies to other descriptions also

Jdallos (approved):

PR #62, opened 2024-12-11
Fixing phase type

### LINKED ISSUE #61: Fixing type for Phase , ResourceAllocation in Schema
Should be changed its type to array in the endpoint instead of in schema

Hanhz2707:


McDuuck (approved):
cool

swagger.yaml
```diff
@@ -2091,39 +2091,41 @@ components:
...
+            format: date
+          deadline:
+            type: string
+            format: date
+          project:
+            type: object
+            description: A reference to Project in Phase Object
+            properties:
+              severaProjectId:
+                type: string
```
dalju005-xamk: format: uuid ?
Also same for severaPhaseId

dalju005-xamk (approved):

swagger.yaml
```diff
@@ -2091,39 +2091,43 @@ components:
...
-              type: string
-              description: Id of the project
-            name:
-              type: string
-              description: Name of the project
-            isClosed:
-              type: boolean
-              default: false
-              description: Boolean value to indicate if the project has been closed or not
+      type: array
```
Jdallos: I think there might be some confusion going on here. As this is the schema I don't see why you need to change it from object to Array. Typically schema is describing an object rather than a list. Rather, where ever it is being referenced e.g. /severa/projects/{severaProjectId}/phases
Should be referenced like:

      responses:
        "200":
          description: A list of phases
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: "#/components/schemas/Phase"
 


Instead of how it is currently:
responses:
        "200":
          description: Phases entry retrieved successfully
          content:
            application/json:
              schema: 
                $ref: "#/components/schemas/Phase"
                
See the list questionnaires endpoint in spec for an example of how it is referring to a schema that is an object but is using it as a list.

If the above makes sense and you agree I guess this also means you will need to make some changes to the resourceAllocations changes you made yesterday.

Jdallos (changes requested):

Jdallos (approved):

PR #69, opened 2025-01-23
Fixing params to query not param for admin sprint view

Hanhz2707:
Now begin to configure admin for sprint view, which need to change params. Instead of severaUserId in path as required value, now , it should only be in query, if no severaUserId (which mean no opt in id user), func will return all allocations (admin privillege)


rimz178 (approved):

naStipe (approved):
Beatiful

Jdallos (approved):

PR #72, opened 2025-02-17
Feature 70 creating api specs for skills for employees

### LINKED ISSUE #70: Creating API specs for skills for employees
I will be adding API specs for employees` skills to be used on a skills page.

naStipe:


swagger.yaml
```diff
@@ -1306,6 +1306,141 @@ paths:
...
+      requestBody:
+        description: Updated user with skills
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UserWithSkills"
+      responses:
+        "200":
+          description: Created a new User with skills
```
```suggestion
          description: Updated a User with skills
```

swagger.yaml
```diff
@@ -1960,4 +2093,42 @@ components:
...
+      required:
+        - name
+        - years
+        - category
+      properties:
+        name:
+          type: string
+          description: Name of the skill
+        years:
+          type: number
```
Jdallos, line 2110: I can see in the existing skill sheet this is a float so I'd suggest doing the same here for now so it's possible for it to handle non whole years.

 Of course it might be ideal for it to be able to cleanly handle months (rather than decimal which does not allow for representing month). How about a year as a number, but also a month as a number, range 0-12?
I think this design would make it easier in the future for this to be further developed to connect skills to projects which will have run for a certain length of time.

swagger.yaml
```diff
@@ -1306,6 +1306,141 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/usersWithSkills":
```
Jdallos: I don't love the name usersWithSkills but I'm not sure of a better one. How about just usersSkills for the endpoints and  UserSkills for the schema?

Jdallos (changes requested):

swagger.yaml
naStipe: Yeah, I agree, that the name does not sound good. I`ll change it according to your suggestion.

swagger.yaml
naStipe: I changed years to months, so that now only the months will be saved and it will be easy to parse it later into years+months, without having to keep two attributes in the database. What do you think of this, is this fine?

Jdallos (approved):

PR #73, opened 2025-02-17
Feature 71 add new api spec entries for user attribute handler

### LINKED ISSUE #71: Add new API spec entries for user attribute handler

rimz178:


swagger.yaml
```diff
@@ -438,19 +439,20 @@ paths:
         - name: id
           in: path
           required: true
-          description: ID of the user whose attribute should be deleted.
+          description: ID of the user whose attribute should be remove.
```
```suggestion
          description: ID of the user whose attribute should be removed.
```

swagger.yaml
```diff
@@ -406,28 +406,29 @@ paths:
           schema:
             type: string
             format: uuid
+        - name: attributeName
+          in: path  
+          required: true
+          description: The name of the attribute to be updated, e.g., `isSeveraOptIn`.
```
```suggestion
          description: The name of the attribute to be updated. Allowed attributes: `isSeveraOptIn`.
```

Jdallos (approved):

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UpdateUserAttribute"
+      responses:
+        "200":
+          description: User attributes retrieved successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UpdateUserAttribute"
```
Jdallos: This describes what the update handler is returning, and currently the update handler is not returning this.
Consider the function is updating both keycloak attributes and severa user id, as such the return value (UpdateUserAttribute) should probably contain all the relevant data. I suppose Attribute contains the severaUserId so I think maybe all that is missing is the  severaKeyword as is returned from the severaAPI- see the comments I left in the lamdba PR.

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+        - name: id
+          in: path
+          required: true
+          description: ID of the user whose attribute should be removed.
+          schema:
+            type: string
+        - name: attributeName
+          in: path  
+          required: true
+          description: The name of the attribute to be removed, e.g., `isSeveraOptIn`.
```
Jdallos: I'm confused, isn't this keyword not attribute? Attribute related to keycloak and is the severaUserId stored there. 

I think the descriptions at least should be updated for the PUT and the delete to make it clear we optIn/ optOut which involves adding/ removing keycloak attributes and severa keywords

Jdallos (approved):

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+      operationId: updateUserAttribute
+      summary: Update user attributes
+      description: Updates a user's attributes based on the provided data.
+      tags:
+        - attribute
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The user attributes to be updated.
```
```suggestion
          description: The id of the user whose attributes are to be updated.
```

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
           description: Invalid request was sent to the server.
 
+  "/users/{id}/attributes/{attributeName}":
+    put:
+      operationId: updateUserAttribute
+      summary: Update user attributes
+      description: Updates a user's attributes based on the provided data.
+      tags:
+        - attribute
+      parameters:
```
Jdallos, line 401: Here you define "attributeName" bute you do not use attributeName from path params in the lamdba.

You use  "attributes" coming from the body (request body provided in this request) which you reference in this spec as being the Attributes schema, but this type does not include "attributeName" or "email" as you are using it in the lambda. So making a request to the lambda based on this spec will not currently work.

In the lambda you should get attributeName from the path param and use accordingly.
You also need to define here where you are getting the email from, it could be another parameter or you could add it to the updateUserAttribute schema. Or perhaps it is coming from the Attributes Schema as you indicate in the lambda? In which case you need to update the schema for Attributes in this spec.

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+      requestBody:
+        description: updated user attributes
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UpdateUserAttribute"
+      responses:
+        "200":
+          description: User attributes retrieved successfully
```
```suggestion
          description: User attributes updated successfully
```

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+          in: path  
+          required: true
+          description: The name of the attribute to be removed from Keycloak and Severa  `isSeveraOptIn`.
+          schema:
+            type: string
+      responses:
+        "204":
+          description: Attribute deleted successfully
+        "default":
+          description: Invalid request was sent to the server.
```
Jdallos: I think this endpoint should be removed from the spec for now.

In the remove-user-attribute lambda you are not actually deleting anything, you are only making PUT requests, so shouldn't this just be an update endpoint and can't this be handled by the already existing update endpoint? Also the remove lambda seems to remove the attribute from the keycloak user but does not remove the keyword from the severa user. The update endpoint currently adds both.



I think for now this can be removed from here. but as additional task if you have time I would suggest removing the remove-user-attribute lambda, and making sure the update-user-attribute can have the update condition when we want to remove the attribute/ keyword. however this does not need to block anything right now as for now we don't need the ability to remove the keyword/ attribute from a user.

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+        - name: id
+          in: path
+          required: true
+          description: ID of the user whose attribute should be removed.
+          schema:
+            type: string
+        - name: attributeName
+          in: path  
+          required: true
+          description: The name of the attribute to be removed from Keycloak and Severa  `isSeveraOptIn`.
```
Jdallos: Seems to me the remove-user attribute lambda is not actually removing anything from severa

swagger.yaml
```diff
@@ -391,6 +391,69 @@ paths:
...
+          schema:
+            type: string
+      responses:
+        "204":
+          description: Attribute deleted successfully
```
Jdallos: The lambda is not deleting anything its only making PUT requests so this is misleading

swagger.yaml
Jdallos: Based on the above it's probably worthwhile investigating the UI as to what you have available to provide to the lambda when making this request e.g. that will tell you if email is in keycloak "Atributes"

swagger.yaml
Jdallos: I unresolved the previous comment I left here because UpdateUserAttribute still does not match what you are returning from the lambda- so this is not correct. 

This should describe what you are returning from this lambda e.g. from line 58 of the updateUserAttributeHandler.

Jdallos (changes requested):

dlama008 (approved):

AYZBTR (approved):

swagger.yaml
```diff
@@ -1457,13 +1432,41 @@ components:
...
+            keycloakResponse:
+              type: object
+              description: The response from the Keycloak update operation.
+              properties:
+                success:
+                  type: boolean
+                updatedFields:
+                  type: array
+                  items:
+                    type: string
```
Jdallos: This indentation needs fixing as it seems to be saying the keycloak response is part of the SeveraUser, but at least in the lambda it's not. For greater clarity I would suggest creating seperate schemas for severaUser and keycloakResponse (same as you have already for Attributes) and then reference them in here as part of the UpdateUserAttributeResponse

swagger.yaml
```diff
@@ -1457,13 +1432,41 @@ components:
...
           description: User's ID in Keycloak.
         attributes:
-          $ref: "#/components/schemas/Attributes"
-        severaUserId:
-          type: string
-          description: The Severa user ID associated with the Keycloak attribute.
-        severaKeyword:
+          $ref: "#/components/schemas/Attributes" 
+
+    UpdateUserAttributeResponse:
```
Jdallos, line 1437: You are missing the "message" attribute here which you are returning form the lambda. I would suggest just removing the message from the lambda as we don't typically return a message. So change required in the spec but message needs to be removed from the lambda updateUserAttributeHandler return.

swagger.yaml
```diff
@@ -1457,13 +1432,41 @@ components:
...
+              description: The Severa user ID associated with the Keycloak attribute.
+            severaKeyword:
+              type: string
+              description: The keyword returned from Severa API that is associated with the user.
+            keycloakResponse:
+              type: object
+              description: The response from the Keycloak update operation.
+              properties:
+                success:
+                  type: boolean
```
Jdallos: Is this really necessary/ used? If not I would suggest removing from here and in the bugfix lambda.

Jdallos (approved):

PR #78, opened 2025-04-07
Thesis Internal Wiki for documentation management

Kseniia-Koshkina:


swagger.yaml
```diff
@@ -1355,6 +1355,221 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/Article"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+  "/articles/{path}":
```
Jdallos: Find endpoints should not accept path. Instead they should use ID, so instead of doing this you should have
list endpoint with query parameter path:
    /articles?path=something
and find endpoint with:
    /articles/{articleId}

Jdallos (changes requested):

swagger.yaml
```diff
@@ -1355,6 +1355,221 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/Article"
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
+    delete:
```
Jdallos, line 1528: We might want some kind of roles here to ensure only "admin" users (check from the accesstoken) can delete and update articles. Also if a user created an article then they could also update/ delete.

This can be changed if it's simple, otherwise this can be left out.

swagger.yaml
```diff
@@ -1355,6 +1355,221 @@ paths:
...
+      responses:
+        "200":
+          description: Updated article entry
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
```
Jdallos: Is this actually needed, haven't you already defined the aticules/id PUT endpoint?

Jdallos (changes requested):

Jdallos (approved):

PR #81, opened 2025-04-14
Update updateUserAttribute requestBody and response structure

balashova12:
- Refactored requestBody to only include `email`.
- `id` and `attributeName` are now passed as path parameters (not in the body).
- Updated response schema to reflect the correct structure.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -397,39 +397,34 @@ paths:
       summary: Update user attributes
       description: Updates a user's attributes based on the provided data.
       tags:
-        - attribute
+        - Users
       parameters:
         - name: id
           in: path
           required: true
-          description: The id of the user whose attributes are to be updated. 
+          description: The keycloak id of the user whose attributes are to be updated.
           schema:
             type: string
             format: uuid
         - name: attributeName
           in: path  
           required: true
-          description: The name of the attribute to be updated. Allowed attributes `isSeveraOptIn`.
+          description: The name of the attribute to be updated.
           schema:
             type: string
       requestBody:
-        description: updated user attributes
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -397,39 +397,34 @@ paths:
...
-          description: The id of the user whose attributes are to be updated. 
+          description: The keycloak id of the user whose attributes are to be updated.
           schema:
             type: string
             format: uuid
         - name: attributeName
           in: path  
           required: true
-          description: The name of the attribute to be updated. Allowed attributes `isSeveraOptIn`.
+          description: The name of the attribute to be updated.
```
```suggestion
          description: The name of the attribute to be updated. Allowed attributes `isSeveraOptIn`.
```

I think the Allowed attributes is helpful at this point to ensure consistency in naming?

swagger.yaml
```diff
@@ -1659,6 +1652,9 @@ components:
           description: URL to the image of the software
         status:
           $ref: "#/components/schemas/SoftwareStatus"
+        review:
+          type: string
+          description: User review of the software
```
Jdallos, line 1657: Any idea why review is showing up in this PR, it seems it is already in the spec develop branch. Di dyou make sure your local branch was up to date before making changes (as in git pulling the develop before creating your branch?).

Just it in case it might be worth running:
git fetch
git merge origin/develop 

on this branch just to make sure it's up to date with the current develop

Jdallos (approved):

swagger.yaml
balashova12: Got it, I’ll do that now

PR #83, opened 2025-04-24
Endpoints Structure

MuhammadKashif064:


dlama008 (approved):

rimz178 (approved):

swagger.yaml
```diff
@@ -8,6 +8,58 @@ security:
   - bearerAuth: []
 
 paths:
+  "/users/{userId}/vacation":
+    put:
+      operationId: updateUserVacation
+      summary: Update user vacation days
+      description: Update vacation days for a specific user by year.
```
```suggestion
      description: Update total and remaining vacation days for a specific user by year.
```

swagger.yaml
```diff
@@ -8,6 +8,58 @@ security:
...
+                type: object
+                properties:
+                  id:
+                    type: string
+                    description: The ID of the updated user
+                  updatedFields:
+                    type: array
+                    items:
+                      type: string
+                    description: List of updated fields (e.g., `vacationDaysByYear`, `unspentVacationDaysByYear`, etc.)
```
Jdallos: Wouldn't it be better to just return the updated attribute and its updated value e.g. vacation days:
{
  total: 30,
  remaining: 15
}

Rather than a list of strings saying what has been updated but not showing how it's been updated.
Think this would be more typical REST conventions

Jdallos: So could just return vacation days

swagger.yaml
```diff
@@ -8,6 +8,58 @@ security:
...
+        "400":
+          description: Invalid request, missing required parameters or data
+        "500":
+          description: Server error while updating vacation days
```
Jdallos: Don't need to specify the different kind of error statuses, just replace this with the error response that is in use for all the other endpoints

swagger.yaml
```diff
@@ -8,6 +8,58 @@ security:
...
+        "200":
+          description: Successfully updated the vacation days
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  id:
+                    type: string
+                    description: The ID of the updated user
```
```suggestion
                    description: The keycloak ID of the updated user
```

swagger.yaml
```diff
@@ -8,6 +8,58 @@ security:
...
+      operationId: updateUserVacation
+      summary: Update user vacation days
+      description: Update vacation days for a specific user by year.
+      tags:
+        - Update vacation
+      parameters:
+        - name: userId
+          in: path
+          required: true
+          description: The ID of the user to update vacation for
```
```suggestion
          description: The keycloak ID of the user to update vacation for
```

Jdallos (changes requested):

swagger.yaml
```diff
@@ -8,6 +8,59 @@ security:
...
+            vacationDays:
+              type: object
+              additionalProperties:
+                type: object
+                properties:
+                  total:
+                    type: integer
+                  remaining:
+                    type: integer
+
```
Jdallos: You should still have the generic error resposne e.g.:
 "default":
          content:
            application/json:
              schema:

swagger.yaml
```diff
@@ -8,6 +8,59 @@ security:
   - bearerAuth: []
 
 paths:
+  "/users/{userId}/vacation":
+    put:
+      operationId: updateUserVacation
+      summary: Update user vacation days
+      description: Update total and remaining vacation days for a specific user by year.
+      tags:
+        - Update vacation
```
Jdallos: I think this should have the Users tag, so it is grouped with the other "Users" endpoints, it should also be moved in the spec to be with the other users endpoints e.g. below /users

swagger.yaml
```diff
@@ -8,6 +8,59 @@ security:
...
+                    properties:
+                      total:
+                        type: integer
+                      remaining:
+                        type: integer
+                  description: Vacation days by year
+              required:
+                - vacationDays
+      responses:
+  "200":
```
Jdallos: Fix indentation, this should be indented within responses

swagger.yaml
```diff
@@ -8,6 +8,59 @@ security:
...
+      application/json;charset=utf-8:
+        schema:
+          type: object
+          properties:
+            id:
+              type: string
+              description: The Keycloak ID of the updated user
+            vacationDays:
+              type: object
+              additionalProperties:
```
Jdallos: This seems  a bit strange to me that we have an object with id and vacationdays, which is an object containing additional properties. Perhaps we need more details here? Why is it called additional properites, is it meant't to be named in such a way to indicate the year it represents? If so this needs to be made clear.

Jdallos: I would suggest making a schema for this (see other schemas in the spec) rather than defining it "inline" here.

Jdallos (changes requested):

swagger.yaml
```diff
@@ -3,6 +3,7 @@ info:
   title: Metatavu Home lambdas
   description: Metatavu Home lambdas
   version: 1.0.0
+
```
Jdallos: No  double line break

```suggestion
```

swagger.yaml
```diff
@@ -460,6 +461,63 @@ paths:
...
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              vacationDays:
+                type: object
+                description: Vacation days by year (e.g., "2024", "2025")
```
Jdallos, line 487: You should describe here that the years are the key for the "additionalProperties" value.
I would also suggest creating a schema to replace additionalProperties and call it something more descriptive e.g. YearlyVacationDays. You can then reference the schema here (as is done with User  inthe 200 response below for example) rather than defining it here.

Jdallos (approved):

PR #84, opened 2025-05-21
Update endpoint name and response

MuhammadKashif064:


AYZBTR (approved):
Approved!

balashova12 (approved):
approved

swagger.yaml
```diff
@@ -502,7 +502,15 @@ paths:
...
             schema:
-              $ref: "#/components/schemas/User"
+              type: object
+              properties:
+                id:
+                  type: string
+                  format: uuid
+                vacationDays:
+                  type: object
+                  additionalProperties:
```
Jdallos, line 512: Isn't the lambda actually returning vacationDays as an object directly, with the attributes total and remaining (the  YearlyVacationDays schema.)

So shouldn't this be:
 properties:
                id:
                  type: string
                  format: uuid
                vacationDays:
                  type: object
                   $ref: "#/components/schemas/YearlyVacationDays"
                   
Note there is no additionalProperties as this attribute/ field does not exist in the lambda.


I think the cleanest way to do this would be to add the id to the YearlyVacationDays schema and then the whole response in the spec would look like this:
 responses:
      "200":
        description: Successfully updated the vacation days
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/YearlyVacationDays"



The schema would be updated to:
 YearlyVacationDays:
      type: object
      description: Vacation allocation and balance for a specific year
      required:
       - id
       - total
       - remaining
      properties:
         id:
          type: string
          format: uuid
        total:
          type: integer
          description: Total vacation days allocated for the year
        remaining:
          type: integer
          description: Remaining vacation days for the year


Does this seem reasonable to you?

Jdallos (approved):

swagger.yaml
Jdallos, line 512: As far as I can tell in the lambda the body contains id and vacation days. VacationDays contains an object with total and remaining. There is no additionalProperties key. So AdditionalProperties should be removed from here. The $ref you have on line 513 is fine but it is the schema for vacationDays, not additionalProperties.

Remove additionaProperties and this is good to go.

Jdallos (approved):

PR #86, opened 2025-06-09
Flextime endpoint update

balashova12:


AYZBTR (approved):

StudVladimir (approved):

MuhammadKashif064 (approved):

swagger.yaml
```diff
@@ -645,20 +645,27 @@ paths:
...
           description: severaId of the user.
           schema:
             type: string
+        - name: eventDate     
+          in: query
+          required: false
+          description: Date for which to get the balance
+          schema:
+            type: string
+            format: date
```
Jdallos: Do we actually need this?

Jdallos: You removed the eventDate in the lambda as it was unused, so I'm not sure why you would add it back in here?

swagger.yaml
```diff
@@ -645,20 +645,27 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
-  "/severa/users/{severaUserId}/flextime":
+  "/users/{userId}/flextime":
```
Jdallos, line 648: This is good, just remember you will need to update the path in the get-flextime-by-user/index.ts in the lambdas

Jdallos (approved):

swagger.yaml
balashova12: I added eventDate to the spec since the Severa API supports it. The lambda doesn’t use it now, but including it keeps the spec flexible in case we want to support custom dates later.

PR #88, opened 2025-06-26
Add /users/flextime endpoint to API specification

AYZBTR:


swagger.yaml
```diff
@@ -1143,7 +1175,7 @@ paths:
                   description: The ID of the Trello card.
                 comment:
                   type: string
-                  description: The comment text.
+                  description: The comment text.:
```
StudVladimir: is this a typo?

StudVladimir (approved):

balashova12 (approved):

swagger.yaml
```diff
@@ -391,6 +391,38 @@ paths:
         "default":
           description: Invalid request was sent to the server.
 
+  "/users/flextime":
```
Jdallos: As this is using data from severa I wonder if it should be structured like /severa/users/flextime.
You will also notice there is already a flextime tag/ endpoint using similar structure (for the flextime for a single user). So it might be good to change the endpoint to match this pattern.

swagger.yaml
```diff
@@ -391,6 +391,38 @@ paths:
...
+    get:
+      operationId: listUsersFlextime
+      summary: Get flextime for multiple users who have opted in
+      description: |
```
Jdallos, line 398: Might be worth noting in this description that this endpoint is (or should be) protected and can only be used by Admin users

swagger.yaml
```diff
@@ -391,6 +391,38 @@ paths:
...
+        Filters users by keywordId to get only opted-in users and returns their flextime balances.
+      tags:
+        - FlexTime
+      parameters:
+        - name: keywordId
+          in: query
+          required: true
+          description: Keyword ID to filter users who have opted in to Severa integration
+          schema:
+            type: string
```
Jdallos: I'm not sure that we should have this parameter, we should only be using this to fetch opted in users and so I'm not sure why it would be needed here.

swagger.yaml
```diff
@@ -1642,6 +1674,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+
```
Jdallos: no need for a double line break

swagger.yaml
```diff
@@ -1689,6 +1722,48 @@ components:
...
+            attributes:
+              type: object
+              description: User's custom attributes
+              properties:
+                severaUserId:
+                  type: string
+                  description: User's ID in Severa system
+                isActive:
+                  type: boolean
+                  description: Whether the user is currently active
```
Jdallos: Do we really need to respond with all this? I think name makes sense in this context (along with Flextime) but that's about it. I'd suggest removing the rest. 
Let me know if you want to discuss further.

Jdallos (changes requested):

swagger.yaml
```diff
@@ -1689,6 +1713,38 @@ components:
...
+            firstName:
+              type: string
+              description: User's first name
+            lastName:
+              type: string
+              description: User's last name
+            email:
+              type: string
+              format: email
+              description: User's email address
```
Jdallos: Do we really need to include email here, again it seems like unnecessary information, we already have this in the UI from the users request that happens on login and I don't see what use it would be here.
As such I'd say remove this also

Jdallos (approved):

PR #92, opened 2025-08-28
Feature-91-Spec-&-Lambda-Clean

jessesillman:
Marked unused parts as deprecated with descriptions and grouped related endpoints.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -14,6 +14,9 @@ paths:
       summary: Lists users avatars
       description: |
         Lists users avatars
+        
+        **Not currently in use. Might be restored in a future version.**
+      deprecated: true
       tags:
         - Slack avatars
       responses:
@@ -26,14 +29,17 @@ paths:
                 items:
                   $ref: "#/components/schemas/Users-avatars"
 
-  "/salesLeads":
+  "/pipedrive/salesLeads":
     get:
       operationId: listSalesLeads
       summary: Lists all Leads from Pipedrive
       description: |
         Lists Leads from Pipedrive
+         
+        **This endpoint is deprecated and will be removed in a future version.**
+      deprecated: true
       tags:
-        - Leads
+        - Pipedrive
       responses:
         "200":
           description: Lists Leads from Pipedrive
@@ -44,14 +50,17 @@ paths:
                 items:
                   $ref: "#/components/schemas/Lead"
 
-  "/salesDeals/{status}":
+  "/pipedrive/salesDeals/{status}":
     get:
       operationId: listSalesDeals
       summary: Lists all Deals from Pipedrive
       description: |
         Lists open Deals from Pipedrive
+         
+        **This endpoint is deprecated and will be removed in a future version.**
+      deprecated: true
       tags:
-        - Deals
+        - Pipedrive
       parameters:
         - name: status
           in: path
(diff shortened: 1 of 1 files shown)


swagger.yaml
```diff
@@ -14,6 +14,9 @@ paths:
       summary: Lists users avatars
       description: |
         Lists users avatars
+        
+        **Not currently in use. Might be restored in a future version.**
+      deprecated: true
```
StudVladimir: All Slack / Trello / Memo are currently marked as deprecated: true. They should not be deprecated, but rather temporarily disabled.

Add a vendor flag, for example:
x-status: inactive
x-deployment: disabled

swagger.yaml
```diff
@@ -1154,6 +1181,10 @@ paths:
       operationId: createComment
       summary: Add a comment to a Trello card
       description: Creates a new comment on a specified Trello card.
+      
+      
+                  **This endpoint is currently not functional and is deprecated.**
```
StudVladimir: If we will safe these endpoints, better to write other description, avoiding work 'deprecated'.
For example: Endpoint not currently deployed; code retained, deployment disabled via serverless config

related to all inactive Slack / Trello / Memo

swagger.yaml
StudVladimir: It's related for all Slack / Trello / Memo endpoints

swagger.yaml
```diff
@@ -1154,6 +1181,10 @@ paths:
       operationId: createComment
       summary: Add a comment to a Trello card
       description: Creates a new comment on a specified Trello card.
+
```
StudVladimir: Extra space, use the same style for description

swagger.yaml
```diff
@@ -1195,11 +1226,15 @@ paths:
       operationId: listCards
       summary: Get all Trello cards on a list
       description: Retrieves a list of cards and their comments from a specified Trello list.
+
```
StudVladimir: Same, Extra space

StudVladimir: Extra spaces in all Trello / Memo endpoints descriptions

StudVladimir (changes requested):
Resolve the comments

swagger.yaml
```diff
@@ -1225,13 +1227,14 @@ paths:
     get:
       operationId: listCards
       summary: Get all Trello cards on a list
-      description: Retrieves a list of cards and their comments from a specified Trello list.
-                   
+      description: |
```
StudVladimir, line 1228: What's the purpose of this sign? Do we need it?

swagger.yaml
```diff
@@ -1217,9 +1255,14 @@ paths:
...
       summary: Create a new Trello card
-      description: Adds a new card to a Trello board.
+      description: |
+        Adds a new card to a Trello board.
+                   
+        **Endpoint not currently deployed; code retained, deployment disabled via serverless config.**
       tags:
         - Trello
+      x-status: inactive
+      x-deployment: disabled
```
StudVladimir: There is probably no need for two vendor flags at once. I put both them as an example of implementation. If we have an inactive status, it is logical that deployment is disabled
You can ask Josh, what is he thinking about it. It's not a mistake, I am just not sure if it's clean

StudVladimir (approved):
Approve with few comments

swagger.yaml
jessesillman: We can use the "|" sign when we want to use multi-line descriptions. Without it, the breaks aren't preserved and the text shows up as a single line, and we don't need to add any extra spaces or empty lines.

JuusoHeikkinen02 (approved):

swagger.yaml
Jdallos: Hmm in some ways I like 2  as we can communicate the status e.g. inactive and if it's deployed or not. But it might be over engineering. Lets just stick with the x-status. Not too familiar with these custom vendor flags myself.

Jdallos (approved):

PR #98, opened 2025-10-03
Feature-97-Adding-New-Fields-In-UserFlextime

jessesillman:
## What has been done

- `email` and `status` fields are now added in the UserFlextime schema to align with the structure of the backend.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1925,14 +1925,9 @@ components:
           type: object
           description: User information
           required:
-            - id
             - firstName
             - lastName
           properties:
-            id:
-              type: string
-              format: uuid
-              description: User's unique identifier
             firstName:
               type: string
               description: User's first name
@@ -2017,7 +2012,7 @@ components:
 
     SeveraUser:
       type: object
-      description: The user's information from Severa.
+      description: The useUserFlextimer's information from Severa.
       properties:
         email:
           type: string
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1939,6 +1939,19 @@ components:
             lastName:
               type: string
               description: User's last name
+            email:
+              type: string
+              description: User's email address
+            attributes:
+              type: object
+              description: User attributes
+              properties:
+                severaUserId:
+                  type: string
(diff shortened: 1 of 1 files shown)


JuusoHeikkinen02 (approved):

swagger.yaml
```diff
@@ -1939,6 +1939,19 @@ components:
             lastName:
               type: string
               description: User's last name
+            email:
```
Jdallos, line 1937: I might be a little out of touch with this feature but can you explain why we need these additions?

Isn't the id we already have the users unique identifier e.g. the severaUserId? If so I don't see why we need attributes at all. Can you find other references to the isActive- I'm not sure that's in use.
As for the email is there a reason we need it as part of the flextime response?

jessesillman: These attributes were hard coded in the employee-flextime-screen.tsx by the previous trainee (PR: https://github.com/Metatavu/metatavu-home/pull/237), and I thought I could add those in the api-spec so that I can simply call the UserFlextime model and fetch the data similarly by calling functions.
For the id, I think it is true that we don't necessarily need that id since the users have their own Severa Id. It should fetch the users' data with their personal SeveraId instead of defining a new one, so it might be redundant indeed. 
IsActive boolean was also defined in both the backend and the frontend by the previous trainee, so that's why I was planning to add that in the api spec too. Here is the ss of the ui with the attributes of the fetched data:
Screenshot from 2025-10-06 09-39-02

Jdallos: Ok, I'd say keep email and isActive then, but maybe we don't need the id duplication?

jessesillman: New commit without the duplicated id (left SeveraUserId).

Alexander-K-byte (approved):

swagger.yaml
Jdallos: Is there a reason severaUserId and isActive are in the attributes object? Can't they just be non nested properties like all the other properties in UserFlextime or does changing this require significant changes somewhere?

Jdallos (approved):

PR #100, opened 2025-10-06
Merging API specs for user skills

naStipe:


swagger.yaml
```diff
@@ -1314,6 +1314,141 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/usersSkills":
+    get:
+      operationId: listUsersWithSkills
+      summary: List Users with skills
+      description: Lists Users with skills
+      tags:
+        - UsersWithSkills
```
Jdallos: Maybe just UsersSkills is better for tag?

Jdallos: If update here then update for all tags

swagger.yaml
```diff
@@ -1314,6 +1314,141 @@ paths:
...
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+    post:
+      operationId: createUserWithSkills
+      summary: Post a new User with skills
+      description: Add new User with skills
```
Jdallos: Again I'm a little confused about the working, if we are creating a UsersSkills entity, then lets just refer to it as UsersSkills rather than as a User with skills- which could lead to confusion

swagger.yaml
```diff
@@ -1314,6 +1314,141 @@ paths:
...
+  "/usersSkills":
+    get:
+      operationId: listUsersWithSkills
+      summary: List Users with skills
+      description: Lists Users with skills
+      tags:
+        - UsersWithSkills
+      responses:
+        "200":
+          description: List of users with skills
```
```suggestion
          description: List of UserSkills
```

swagger.yaml
Jdallos: I would be consistant and use UserSkills as the name of the entity schema and the endpoint

swagger.yaml
```diff
@@ -2025,6 +2160,43 @@ components:
...
+    UserSkills:
+      type: object
+      description: A user entity with skills attribute
+      required:
+        - name
+      properties:
+        id:
+          type: string
+          format: uuid
+          description: ID of a user with skills attribute from keycloak
```
Jdallos: So is this the users keycloak Id? If so just say a users keycloak ID.

swagger.yaml
```diff
@@ -2025,6 +2160,43 @@ components:
...
+          default: null
+        name:
+          type: string
+          description: Name of the user with skills attribute
```
Jdallos: Is this the users full name?

Jdallos (changes requested):

Jdallos (approved):

PR #103, opened 2025-10-30
feature-102-Get-users-work-week-for-vacations

Alexander-K-byte:


swagger.yaml
```diff
@@ -923,6 +923,47 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/severa/users/{severaUserId}/workWeek":
+    get:
+      operationId: getUserContractedWorkWeek
```
Jdallos: This could be just:
getUserWorkWeekLength? Maybe clearer and maybe the contracted is unnecessary- this could be true in other places if you agree?

swagger.yaml
```diff
@@ -2085,6 +2126,24 @@ components:
...
+            type: integer
+            minimum: 0
+            maximum: 6
+        contractedDaysCount:
+          type: integer
+          description: Total number of contracted workdays in the week
+      required:
+        - contractedWeek
+        - contractedDaysCount
+
```
Jdallos: So as previous comment I suggested UserWorkWeekLength, let me know what you think.

Also do we really need to return this contractedWeek data? If so why?
 Can't we jsut return the number of days (contractedDayCount, which I would perhaps rename workingWeekLength).


I think typically the required is put before the properties for clarity.

swagger.yaml
Alexander-K-byte: Yeah, makes sense

swagger.yaml
Alexander-K-byte: the contractedWeek returns the array for pt/ft workers, lets us accurately calculate if for example you requested Wed - Sun in request, it will correctly only remove 2 days instead of 3 since the array can be used to show you only work wed/thu in that date range.

Jdallos (approved):

swagger.yaml
```diff
@@ -939,11 +939,11 @@ paths:
             type: string
       responses:
         "200":
-          description: The user's contracted work week.
+          description: The user's contracted work week as a numbered array and total work days.
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/getUserWorkWeekLength"
+                $ref: "#/components/schemas/getUserContractedWorkWeek"
```
Jdallos: Name needs updating as you have renamed the schema to : calculateUserContractedWeek

Alexander-K-byte: Yip, forgot about the ref

PR #105, opened 2025-11-03
feature-104-lambdas-handler-for-workdays-chart

JuusoHeikkinen02:
Created new endpoint for fetching users workdays for a specific time frame.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -583,11 +583,11 @@ paths:
                 
   "/severa/users/{severaUserId}/workdays":
     get:
-      operationId: getWorkdaysForUser
-      summary: Get user's workdays for a specific time period
-      description: >
-        Fetches a user's workdays from Severa for the given date range.
-        Each returned record includes data about expected vs. entered hours, absences, overtime permissions, and time entry completeness.
+      operationId: listWorkdaysForUser
+      summary: Lists user's workdays for a specific time period
+      description: 
+        Lists a user's workdays from Severa for the given date range.
+        Each returned record includes data about the date, user GUID, expected hours, entered hours, and whether the day is a holiday.
       tags:
         - Severa
       parameters:
@@ -619,7 +619,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "#/components/schemas/getWorkdaysForUser"
+                  $ref: "#/components/schemas/listWorkdaysForUser"
         "401":
           description: Missing or invalid Severa user ID.
           content:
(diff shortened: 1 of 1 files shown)


MarcelMeta (approved):

Alexander-K-byte (approved):

swagger.yaml
```diff
@@ -580,6 +580,64 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+                
+  "/severa/users/{severaUserId}/workdays":
+    get:
+      operationId: getWorkdaysForUser
+      summary: Get user's workdays for a specific time period
+      description: >
```
```suggestion
      description:
```

swagger.yaml
```diff
@@ -580,6 +580,64 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+                
+  "/severa/users/{severaUserId}/workdays":
+    get:
+      operationId: getWorkdaysForUser
```
Jdallos: Rather than getWorkdaysForUser or Get user's workdays for a specific time period in the spec lets be explicit if we are finding or listing, so in this case listWorkdaysForUser and List user's workdays....

That should match the majority of the already exisitng endpoint in the spec hopefully

swagger.yaml
```diff
@@ -580,6 +580,64 @@ paths:
...
+      summary: Get user's workdays for a specific time period
+      description: >
+        Fetches a user's workdays from Severa for the given date range.
```
Jdallos: What is no dates are provided, does it return it for all time?

swagger.yaml
```diff
@@ -2609,6 +2667,72 @@ components:
...
+      required:
+      - date
+      - userGuid
+      - expectedHours
+      - enteredHours
+      - absenceHours
+      - incompleteTimingMinutes
+      - enteredTimeEntries
+      description: Workday details for a specific user
+      properties:
```
Jdallos, line 2679: These all look fine but we should also make sure we don't process more data than necessary. I would suggest checking with Marcel, if any of these properties are not actually going to be used in the front end then they should be removed from the spec and the lambda should map over the severa API response to ensure we do not send any unnecessary data.

Jdallos (approved):

swagger.yaml
JuusoHeikkinen02: Dates are required for the severa api so it will not return all time. But im going to call the service that uses default values of a week if dates are not provided.

PR #108, opened 2025-11-05
feature-106-get-oncall-user-email

JuusoHeikkinen02:
created new endpoint for getting users email by current on call user's username

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -424,6 +424,39 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/on-call/users/{username}":
+      get:
+        operationId: getOnCallUserEmail
+        summary: Get on call user's email by username
+        description: Retrieves on call user's email by current week's on call username.
+        tags:
+          - On call
+        parameters:
+          - name: username
+            in: path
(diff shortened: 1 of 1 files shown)


Alexander-K-byte (approved):

jessesillman (approved):

swagger.yaml
```diff
@@ -2417,6 +2450,18 @@ components:
...
+    
+    OnCallUser:
+      type: object
+      description: Minimal user object containing only the email of the on-call user.
+      properties:
+        email:
+          type: string
+          format: email
+          description: The email address of the on-call user.
+          example: "alex.jones@example.com"
```
Jdallos, line 2462: Some might consider this an unnecessarily controversial example.

swagger.yaml
```diff
@@ -424,6 +424,39 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/on-call/users/{username}":
```
Jdallos, line 427: Do we really need an enpoint and entire lambda for this? Isn't this something that could be done in the existing on call fetch/ listing?

PR #112, opened 2025-11-11
Feature-110-Severa-Opt-Out

jessesillman:
## What has been done
- Added the **DELETE** endpoint to  _/users/{userId}/severa-opt-in_ endpoint to enable Severa opt-out toggle.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -708,6 +708,39 @@ paths:
         "default":
           description: Invalid request was sent to the server.
 
+  "/users/{userId}/severa-opt-in":
+    delete:
+      operationId: removeSeveraOptIn
+      summary: Remove Severa opt-in for user
+      tags:
+        - Users
+      parameters:
+        - name: userId
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+          description: The Keycloak ID of the user to remove Severa opt-in for
+      security:
+        - bearerAuth: []
+      responses:
+        "200":
+          description: Opt-out completed
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  message:
+                    type: string
+        "default":
+          description: Invalid request was sent to the server.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+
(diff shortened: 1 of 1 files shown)


JuusoHeikkinen02 (approved):

Alexander-K-byte (approved):

swagger.yaml
```diff
@@ -708,6 +708,40 @@ paths:
         "default":
           description: Invalid request was sent to the server.
 
+  "/users/{userId}/severa-opt-in":
```
Jdallos, line 711: Great I prefer this endpoing to how it's being done when opting in e.g.   "/users/{id}/attributes/{attributeName}"

I don't want to add to this feature you are currently working on but I would prefer if the opt in endpoint could also be refactored/ simplified in this way to be specific rather than the generic way it is currently.

I'd suggest creating a trello task for this as a future task and taking note of the changes you would make to the spec and the lambda for opting in- it no longer needs to handle attributes for example, as it would only be used for the single purpose of opting in.

swagger.yaml
```diff
@@ -708,6 +708,40 @@ paths:
...
+    delete:
+      operationId: removeSeveraOptIn
+      summary: Remove Severa opt-in for user
+      description: Remove the user's Severa opt-in status.
+      tags:
+        - Users
+      parameters:
+        - name: userId
```
Jdallos, line 719: In severa we look up the user by email, I assume the keycloak Id is enough to allow us to look up the user email in the lambda? This would be preferred to including the email here if possible. And email look up form keycloak id is possible, then removing the email from the update attribute endpoint would also be an improvement to be done in the future task I mentioned in the previous comment

Jdallos (approved):

swagger.yaml
jessesillman: Roger, I will create a trello card with the notes of this for the future task! :)

swagger.yaml
jessesillman: Yes, the lambda can look up the user email with the keycloak id. I will add this to the trello card and note removing the email from the update attribute endpoint.

PR #115, opened 2025-11-24
Feature-114-Refactor-Opt-In-Endpoints

jessesillman:
## What has been done
- Added **PUT** _/users/{userId}/severa-opt-in_ to expose opt-in as a new endpoint.
- Removed the email field from the requestBody of _/users/{id}/attributes/{attributeName}_, requiring only value.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -639,50 +639,6 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
-  "/users/{id}/attributes/{attributeName}":
-    put:
-      operationId: updateUserAttribute
-      summary: Update user attributes
-      description: Updates a user's attributes based on the provided data.
-      tags:
-        - Users
-      parameters:
-        - name: id
-          in: path
-          required: true
-          description: The keycloak id of the user whose attributes are to be updated.
-          schema:
-            type: string
-            format: uuid
-        - name: attributeName
-          in: path
-          required: true
-          description: The name of the attribute to be updated. Allowed attributes `isSeveraOptIn`.
-          schema:
-            type: string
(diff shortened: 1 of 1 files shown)


Alexander-K-byte (approved):

Dshad-Anon (approved):

MariaKatariinaH (approved):

swagger.yaml
```diff
@@ -661,18 +661,18 @@ paths:
...
-        description: Email of the user whose attribute is being updated.
+        description: Attribute update payload.
         required: true
         content:
           application/json:
             schema:
               type: object
               required:
-                - email
+                - value
```
Jdallos: What is this value? I thought the point of this refactor was to make the endpoint more specific so rather than the user providing some generic/ flexible value e.g. SeveraOptIn string it only dealys with severa opt in?
Do we no longer need to pass the email here?

swagger.yaml
```diff
@@ -709,6 +709,38 @@ paths:
           description: Invalid request was sent to the server.
 
   "/users/{userId}/severa-opt-in":
+    put:
+      operationId: addSeveraOptIn
+      summary: Add Severa opt-in for user
+      description: Add the user's Severa opt-in status.
```
```suggestion
      description: Add the user's Severa opt-in status to both severa and keycloak.
```

swagger.yaml
Jdallos: Ok so as you have the severa opt in put below, can't this endpoint now just be removed?

jessesillman: You are right, we don't really need it so it will be removed in the next commit.

Jdallos (approved):

PR #117, opened 2025-12-08
Restore-116-management-memos-endpoints

JuusoHeikkinen02:
Restored endpoints related to management memos and created a new endpoint for creating/storing translated memo entries to dynamoDB

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -1854,7 +1854,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
-  "/google-drive/memo-translate/{id}/{language}":
+  "/translate-memo/{id}/{language}":
     get:
       operationId: getTranslatedMemoPdf
       summary: Get translated memo content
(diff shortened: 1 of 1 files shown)


Alexander-K-byte (approved):

Dshad-Anon (approved):

swagger.yaml
```diff
@@ -1862,17 +1854,57 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
-  "/google-drive/memo-translate":
+  "/google-drive/memo-translate/{id}/{language}":
```
Jdallos: So if this endpoint "gets" the translated memo, isn't this actually something that should be got from the dynamoDB? If so seems confusing its under googel-drive endpoint?
-I guess there will be the need for some endpoint to get the original memo and translate/ summarize but imagine that should be under the name of the 3rd party service tool, or maybe better translate/

swagger.yaml
```diff
@@ -1862,17 +1854,57 @@ paths:
...
+            schema:
+              $ref: "#/components/schemas/Error"
+      "500":
+        description: Internal server error
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Error"
+
+  "/google-drive/memo-translate":
```
Jdallos: Ok so here is the create translations memo, but should it really be under google-drive?
What about translate-memo/

Jdallos (approved):

PR #121, opened 2025-12-17
Restored the get slack avatar endpoint

JuusoHeikkinen02:
Restored the get slack avatar endpoint and made the endpoint fetch only the slack avatar for individual user by their email.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -8,26 +8,27 @@ security:
   - bearerAuth: []
 
 paths:
-  "/slack-avatar":
+  "/slack-avatar/{email}":
     get:
-      operationId: slackAvatar
-      summary: Lists users avatars
-      description: |
-        Lists users avatars
-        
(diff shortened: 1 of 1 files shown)


Dshad-Anon (approved):

jessesillman (approved):

swagger.yaml
```diff
@@ -8,26 +8,27 @@ security:
   - bearerAuth: []
 
 paths:
-  "/slack-avatar":
+  "/slack-avatar/{email}":
```
Jdallos: As said in the lambda review I think email should probably be query param rather than path- do you agree?

Jdallos (changes requested):

swagger.yaml
JuusoHeikkinen02: yes will change it to use query param

Jdallos (approved):

PR #134, opened 2026-03-19
add user status endpoint

rashinikarunarathne:


Dshad-Anon (approved):

MikaelPL1 (approved):

swagger.yaml
Jdallos, line 1: Please note the issue number is missing from the branch name. Please check the documentation in trello under creating your first PR for branch name conventions. No need to change this one but please name accordingly in the future

swagger.yaml
```diff
@@ -824,6 +824,56 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/severa/users/{severaUserId}/status":
+    put:
+      operationId: updateUserStatus
+      summary: Update user active status
+      description: Updates the isActive attribute of a user in Keycloak using the Severa user ID.
```
Jdallos: If this is updating the isActive attribute in the keyclaok then this endpoint should be grouped with the other keycloak editing endpoints in this spec e.g. /Users- which gets a list of keycloak users, or users/usersId which searches for a keycloak user using the keycloak id.

swagger.yaml
```diff
@@ -824,6 +824,56 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
 
+  "/severa/users/{severaUserId}/status":
```
```suggestion
  "/users/{UserId}/status":
```

Jdallos: See comments below to explain this change

swagger.yaml
Jdallos: This has nothing to do with the severa data, so while keycloak user _should_ contain a severa id and you _could_ try to find the keycloak user via severa id , this is **not** the typical practice.
You should search for a given entity by it's associated id e.g. you are looking for a keycloak user (to update its status) so you should find the user via the keycloak id. Similar to the users/usersId endpoint that already exists in this spec.


So this description will need to be updated accordingly

swagger.yaml
```diff
@@ -824,6 +824,56 @@ paths:
...
+    put:
+      operationId: updateUserStatus
+      summary: Update user active status
+      description: Updates the isActive attribute of a user in Keycloak using the Severa user ID.
+      tags:
+        - Users
+      parameters:
+        - name: severaUserId
```
Jdallos: As said in other comments this should be keycloakId- unless there is some reason you are using this that I'm missing?

Jdallos (changes requested):

Jdallos (approved):

PR #141, opened 2026-04-08
add Slack user away status update endpoint and request/response schemas

arodriguezvaz:


rashinikarunarathne (approved):

swagger.yaml
```diff
@@ -2409,6 +2407,48 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+
+  "/slack/set-user-away":
```
Jdallos: Can we improve the endpoint name?
check-vacation-status.
Also maybe set-user-away is ok if you think so

swagger.yaml
```diff
@@ -2409,6 +2407,48 @@ paths:
...
+    post:
+      operationId: setSlackUserAway
+      summary: Set Slack user as away during vacation
+      description: Saves a request to update a user's Slack profile status when vacation starts. When the startDate arrives, the backend automatically updates the user's status in Slack with emoji and status text, with automatic expiration on the vacation end date. No notifications are sent to channels.
```
Jdallos: If you think this will be some kind of function running on a daily timer (a cron job) it should be detailed briefly here (how often it will run)

swagger.yaml
```diff
@@ -3493,4 +3533,57 @@ components:
...
+          type: string
+          format: date
+          description: End date of the vacation in YYYY-MM-DD format. Slack automatically clears the status at this time.
+          example: "2026-04-22"
+
+    SlackUserAwayResponse:
+      type: object
+      description: Response confirming that the Slack away status request has been saved. The actual Slack profile update will occur on the startDate with automatic expiration on endDate.
+      properties:
+        saved:
```
Jdallos: I'm unclear on this, the request has been saved refers to what. Won't this function just run each day and if the start or end date is today it will make the slack request? In that case we might want some "saved" or "updated" response with number of responses

swagger.yaml
```diff
@@ -1307,6 +1307,19 @@ paths:
           description: Filter Vacation Requests by draft state
           schema:
             type: boolean
+        - name: startsOn
+          required: false
+          in: query
+          description: Return vacation requests that start on the given date (YYYY-MM-DD).
+          schema:
+            type: string
+            format: date
```
Jdallos: Couldn't you use the already existing before and after query params to do this? I think you should be able to and so this addition is unnecessary.

Jdallos (approved):

swagger.yaml
```diff
@@ -2415,6 +2419,37 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+
+  "/jobs/vacation-slack-sync":
+    post:
+      operationId: vacationSlackSync
+      summary: Checks for vacations starting at the current time and updates Slack status
+      description: Internal job triggered by a scheduler based on a configured cron expression.
```
Jdallos, line 2427: State here how often it will run (I'd suggest daily or Mon-Fri)

swagger.yaml
```diff
@@ -3498,4 +3533,20 @@ components:
...
\ No newline at end of file
+          example: medium
+
+    VacationSlackSyncResult:
+      type: object
+      required:
+        - processedUsers
+        - startedVacations
+        - endedVacations
+      properties:
```
Jdallos, line 3544: Aren't properties missing the endedVacations that you say is required?

swagger.yaml
```diff
@@ -2415,6 +2419,37 @@ paths:
...
+      operationId: vacationSlackSync
+      summary: Checks for vacations starting at the current time and updates Slack status
+      description: Internal job triggered by a scheduler based on a configured cron expression.
+        This job runs periodically and is responsible for checking vacation data
+        and synchronizing Slack user statuses accordingly.
+
+        The job evaluates which vacations start  at the current time and updates Slack status for users whose vacation is starting
+
+        This endpoint is documented for specification completeness and
+        is not intended to be invoked manually.
```
Jdallos, line 2434: Good

Jdallos (changes requested):

swagger.yaml
```diff
@@ -2415,6 +2419,37 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+
+  "/jobs/vacation-slack-sync":
+    post:
+      operationId: vacationSlackSync
+      summary: Checks for vacations starting at the current time and updates Slack status
+      description: Internal job triggered by a scheduler based on a configured cron expression.
+        This job runs dailly and is responsible for checking vacation data
```
```suggestion
        This job runs daily and is responsible for checking vacation data
```

Jdallos (approved):

PR #146, opened 2026-04-19
spec update

rashinikarunarathne:
This change introduces role-based personalization to the system.
Users can now select their job roles (e.g. developer, designer), which are stored in Keycloak. These roles are then used to personalize different parts of the application.
Software recommendations can be filtered based on roles
Wiki articles can be targeted to specific roles
Questionnaires can be suggested based on user roles
Overall, this allows the system to provide a more relevant and user-specific experience instead of showing the same content to everyone.

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -976,6 +976,24 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/UserRolesResponse"
+        "400":
+          description: Invalid user ID supplied
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "404":
+          description: User not found
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref:"#/components/schemas/Error"
         
     put:
       operationId: updateUserRoles
@@ -3661,11 +3679,3 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Role"
-
-    SuccessMessage:
-      type: object
-      properties:
-        message:
-          type: string
-          example: Roles updated successfully      
-  
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -950,7 +950,153 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Error"              
+                $ref: "#/components/schemas/Error"
+                
+   "/users/{userId}/roles":
+    get:
+      operationId: getUserRoles
+      summary: Get user roles
+      description: Retrieve roles assigned to a user from Keycloak
+      tags:
+        - Users
+      parameters:
+        - name: userId
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
+        "200":
+          description: List of roles assigned to the user
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+                  example: developer
+
+    put:
+      operationId: updateUserRoles
+      summary: Update user roles
+      description: |
+        Allows users to self-select roles. Updates roles in Keycloak. Restricted roles (e.g. admin) cannot be assigned by users.
+      tags:
+        - Users
+      parameters:
+        - name: userId
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - roles
+              properties:
+                roles:
+                  type: array
+                  items:
+                    type: string
+                  example:
+                    - developer
+                    - designer
+      responses:
+        "200":
+          description: Roles updated successfully
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  message:
+                    type: string
+                    example: Roles updated successfully
+        "400":
+          description: Invalid roles provided
+        "403":
+          description: Attempt to assign restricted roles
+
+  "/software/recommended":
+    get:
+      operationId: getRecommendedSoftware
+      summary: Get recommended software
+      description: Returns software filtered based on user roles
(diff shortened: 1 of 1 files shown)


arodriguezvaz (approved):

NeonValtokari (approved):

swagger.yaml
```diff
@@ -950,7 +950,153 @@ paths:
...
+          description: List of roles assigned to the user
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+                  example: developer
+
+    put:
```
Jdallos, line 998: You can double check the indentation and formatting of the swagger is correct by pasting the file into the https://editor.swagger.io/ Please check formatting there before merging

swagger.yaml
```diff
@@ -950,7 +950,153 @@ paths:
...
+              type: string
+      responses:
+        "200":
+          description: Filtered questionnaires
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Questionnaire"
```
Jdallos: Is this part of the task, I thought you were focusing on a user being able to be assigned roles. I think recommendations based on the role should maybe be done as further work in a seperate feature.


I think rather than have new endpoints we might want to integrate a query param into the existing software/questionnaire/article endpoint to allow us to filter by given tags. Probably then in the UI we would call the endpoint to list all (as it does already) and also call it to provide the recommended based on role entities. This would then need some spec update, lambda update to add the filtering and UI update to have a new recommended section.
You are likely a good person to do this task in the future, I'll add these notes to trello and allocate you. 
But for now remove these/recommended endpoints

Jdallos: https://trello.com/c/zL98d3GT/490-recommended-software-articles-questionnaires

swagger.yaml
```diff
@@ -3553,4 +3699,30 @@ components:
...
+          description: Roles this software is recommended for
+
+  Questionnaire:
+      type: object
+      properties:
+        roles:
+          type: array
+          items:
+            type: string
+          description: Roles this questionnaire is intended for
```
Jdallos: So this can also be removed for now

Jdallos (approved):

swagger.yaml
```diff
@@ -952,31 +962,22 @@ paths:
...
                 
-   "/users/{userId}/roles":
+  "/users/{userId}/roles":
     get:
       operationId: getUserRoles
       summary: Get user roles
       description: Retrieve roles assigned to a user from Keycloak
-      tags:
-        - Users
+      tags: [Users]
```
Jdallos: Any reason for this change, other tags in the spec are defined how this was previously e.g. 
tags:
        - Users

Jdallos (approved):

swagger.yaml
```diff
@@ -959,6 +1012,16 @@ paths:
       description: Retrieve a list of all questionnaires in the registry.
       tags:
         - Questionnaires
+      parameters:
+        - name: roles
+          in: query
+          required: false
+          style: form
+          explode: true
```
Jdallos: We haven't used the style and explode attrbute in any of the other endpoints in this spec. I belive the reason for this is because form and true and the default values. Please check and remove if you agree- this helps us avoid bloat in the spec

swagger.yaml
```diff
@@ -2114,6 +2177,15 @@ paths:
           description: Get draft articles.
           schema:
             type: boolean
+        - name: roles  
+          in: query
+          required: false
+          style: form
+          explode: true
```
Jdallos: We haven't used the style and explode attrbute in any of the other endpoints in this spec. I belive the reason for this is because form and true and the default values. Please check and remove if you agree- this helps us avoid bloat in the spec

swagger.yaml
```diff
@@ -2457,6 +2529,15 @@ components:
...
 
+  parameters:
+  UserId:
+    name: userId
+    in: path
+    required: true
+    schema:
+      type: string
+      format: uuid  
+
```
Jdallos, line 2553: This indentation seems wrong, I'm also not sure this is needed and could be just defined in line, which might be clearer and more in keeping with existing spec format?

Additionally if you paste this into the swagger editor you will see there are multiple errors in the file- so it needs changes.

Jdallos (changes requested):

swagger.yaml
```diff
@@ -950,7 +958,51 @@ paths:
...
+        - Users
+      parameters:
+        - $ref: "#/components/parameters/UserId"
+      responses:
+        "200":
+          description: List of roles assigned to the user
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/UserRolesResponse"
```
Jdallos, line 978: Missing error response like most other endpoints have?
Might be worth adding them in at this point.

swagger.yaml
```diff
@@ -3553,4 +3629,43 @@ components:
...
+
+    Role:  
+      type: string
+      enum:
+        - developer
+        - designer
+        - architect
+        - admin
+        - management
+        - trainee
```
Jdallos, line 3660: Would this seems fine for now, I wonder if in the future it would be possible/ better to have the list of roles as open, or at least so we can update them. But again this is quite fine for first version

swagger.yaml
```diff
@@ -3553,4 +3629,43 @@ components:
...
+      type: object
+      required:
+        - roles
+      properties:
+        roles:
+          type: array
+          items:
+            $ref: "#/components/schemas/Role"
+
+    SuccessMessage:
```
Jdallos: Don't think we have done this in any other endpoints, is it really needed here? I'd suggest removing.

swagger.yaml
Jdallos: Also should we update the User endpoint and response schema so it also returns the user role? AGain this could be further work.

Jdallos (approved):

Jdallos (approved):

